PR #3, opened 2015-12-10
Change header order so setup.py works on OS X.

sybenzvi:
Hi Giacomo et al., I had to change the order of the includes in threeML/models/pyToCppModelInterface.h in order to get setup.py to build on OS X.  You may want to copy the change to the master.


giacomov:
Thank you!
Colas and I spent some time on the workshop trying to figure this out, but we couldn't. 

I am going to merge this in the main trunk.

PR #12, opened 2016-03-08
nTransits calculation for HAWCLike

cdr397:
Omitting the nTransit when creating HAWCLike object should allow the system to calculate (same as how it is done in LiFF) its own nTransits and use it.

### DIFF
--- a/threeML/plugins/HAWCLike.py
+++ b/threeML/plugins/HAWCLike.py
@@ -15,6 +15,8 @@ from matplotlib import gridspec
 
 import numpy
 
+import ROOT
+
 defaultMinChannel = 0
 defaultMaxChannel = 9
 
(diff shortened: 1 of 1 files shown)


giacomov:
For reference, I report here a conversation we had on another channel:

Instead of reading the map from the 3ML side only to get the number of transits (which also add a dependency on ROOT), we should use the constructor for HAWCLike which compute the number of transit by itself. This way the reading of the map would be done only once, and we would avoid to add the dependence on ROOT. Could you do that instead?

@giacomov Sure, that sounds much better! I basically applied the same method as how LiFF was finding the nTransits, I didn't like the fact that I was adding a dependency on ROOT either. I'll try that instead. Thanks.

I am going to close this pull request now.

PR #46, opened 2016-06-23
Added check in is_module_available for lowercase modules... like on OS X :)

drJfunk:
There is just an extra loop in the function. pyOpt = pyopt on OS X

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,15 +16,24 @@ execfile('threeML/version.py')
 def is_module_available(module_name):
 
     # Fast path: see if the module has already been imported.
-
+    # Also check if the module is lower case named on some systems
     try:
 
         exec('import %s' % module_name)
 
     except ImportError:
 
-        return False
+        try:
+            
+            exec('import %s' % module_name.lower())
 
+        except ImportError:
+
+            return False
+
+        else:
+
+            return True
     else:
 
         return True


giacomov:
pyopt and pyOpt are two different things, unfortunately :-)

The first one is a command line utility:

https://pypi.python.org/pypi/pyopt

while the second one is a suite of optimizers:

http://www.pyopt.org/

Thus the case-sensitive check is actually the right thing to do!

drJfunk:
doh! Ok, I’ll check this out!


> On Jun 27, 2016, at 7:47 AM, Giacomo Vianello <notifications@github.com> wrote:
> 
> pyopt and pyOpt are two different things, unfortunately :-)
> 
> The first one is a command line utility:
> 
> https://pypi.python.org/pypi/pyopt <https://pypi.python.org/pypi/pyopt>
> while the second one is a suite of optimizers:
> 
> http://www.pyopt.org/ <http://www.pyopt.org/>
> Thus the case-sensitive check is actually the right thing to do!
> 
> —
> You are receiving this because you were assigned.
> Reply to this email directly, view it on GitHub <https://github.com/giacomov/3ML/pull/46#issuecomment-228659013>, or mute the thread <https://github.com/notifications/unsubscribe/AGc157MSmYhcc2br38ztCoT9XVcULk8_ks5qP2PngaJpZM4I8f33>.
> 

+++++++++++++++++++++++++++++++
J Michael Burgess
+46 72-381 30 20
jmichaelburgess@gmail.com

Issue #51, opened 2016-08-29
Jointlikehood display


drJfunk:
With the migration to pandas, should we abandon the custom tables in the JL display so that all the data (matrix, MLE, etc) are pandas? If so, I would like to make them a little nicer on the eyes with pandas style. Thoughts?

giacomov:
See #52

PR #52, opened 2016-08-30
Pandas style

drJfunk:
In response to #51 I have added pandas styling to the joint likelihood and bayesian outputs for visual clarity. There are a few functions added to rich_display.  The functions for displaying may need to be tweaked and can apply to other output (astromodels?)

Some of the Table classes may be obsolete now that we are using pandas.

### DIFF
--- a/threeML/bayesian/bayesian_analysis.py
+++ b/threeML/bayesian/bayesian_analysis.py
@@ -28,7 +28,7 @@ from threeML.config.config import threeML_config
 from threeML.io.progress_bar import progress_bar
 from corner import corner
 from threeML.exceptions.custom_exceptions import LikelihoodIsInfinite, custom_warnings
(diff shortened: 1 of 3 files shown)


drJfunk:
screenshot 2016-08-30 11 16 46

giacomov:
I have merge the styler in the rich_display, but we refrain from using it as the style functionality in pandas is still in development and experimental according to the documentation.

For example, it does not allow the display of frames in terminals, which would not allow to use 3ML in scripts.

In the future, when it will be more stable, we will be able to re-implement it quickly.

grburgess:
It looks like they have made it official. We can prettifiy 3ML now.
https://pandas.pydata.org/pandas-docs/stable/style.html

Just like _Nature_, analysis is all about the look.

PR #56, opened 2016-09-05
Update tte like

drJfunk:
This branch addresses the following:

 Adds general EventList class capable of producing a PHA type readable by the new OGIP plugins.

Updates the FermiGBMTTELike class to the new OGIP plugin.

Modifies the PHA class to be compliant with the above changes.

Adds a general PHAContainer class that can be used to build PHAs on the fly.

Updates the Response class to handle GBM response with no TLMIN key work (see #53 )

Adds a stats_tool file for simple things like Li and Ma significance

Moved plotting features from original ogip.py file to their appropriate places.

Enables multiple source selection in TTE files.

--
Examples have been updated accordingly


drJfunk:
I need to check a bug!

PR #61, opened 2016-09-08
Adding SpectralFlux class

drJfunk:
This adds a class which can do flux calculations with proper error propagation.
There is a new example, Flux_Calculations, which demonstrates the usage.

There is additionally, a small change in the TTELike class which adds some additional print out info.


giacomov:
This is very nice!

However, the best way to compute the flux error for the MLE case is using the integral of the function as a parameter of the fit, so that the error on the flux can be asymmetric as it is computed as usual as the range which gives a likelihood of (best fit likelihood) +/- 0.5 (for a 1-sigma error). This method gets rid of the assumptions of the error propagation, which are: 1) symmetric errors, 2) linear correlations among parameters (i.e., the contour between any two parameters is a 2d Gaussian). 

Implementing that is straightforward for a single component (I could write down the math, it's pretty easy), but it requires a little more thoughts for a multicomponent model.

drJfunk:
Yes. This method is a bit general. I think it is good to have option of both perhaps? One can always use the flux as a parameter. This is similar to e.g., fitting the background with source to "properly" propagate the errors. I think we could keep these separated.  In reality, there is not a method to asymmetrically propagate errors unless you make a grid or use Bayesian sampling. 

The other advantage of "post-fit" flux calculation with this method, is that you can recalculate the parameter ranges on the fly. If one wants the asymmetric errors, then they could build a function that includes the flux (even for components).

giacomov:
I'm merging the pull request right now, this is all very nice.

Let's keep thinking about how to implement the functionality I was talking about, which would probably be part of this new class you made.

PR #65, opened 2016-09-20
Ogip model plot and fixes to TTELike + OGIPLike

drJfunk:
This adds a data plot function to OGIPLike. For the moment, it is a separate function due to the way models and over plotting is handled. 

Also fixes a few bugs in EventList.

Adds an "multi mission" time look up to EventList if there is a web connects

Modifies Rebinner to handle energy selection.

Several small fixes and renames. 

Demo updated to show features.

Should be last commit before we add tests!


threeML/plugins/OGIPLike.py
```diff
@@ -72,19 +79,97 @@ def rebin(self, *vectors):
...
+        return self._min_counts
+
+
+
+    @property
+    def edges(self):
+
+        return self._edges
+
+    @edges.getter
```
giacomov: Isn't this useless since edges is already a property?

drJfunk:
It is to format how the getting outputs. I can have a look at the google format and change if needed.

PR #66, opened 2016-09-22
Minor updates to OGIP related files

drJfunk:
Moved property decorators to follow Google style guide. 

Major code cleanup for safety.

Implemented parallel per channel background fitting in EventList to increase speed. (~factor 3x for GBM TTE)

XRT example is still suffering from XSPEC models


threeML/plugins/OGIP/eventlist.py
```diff
@@ -388,190 +408,225 @@ def _fit_background(self):
...
 
-            self.optimalPolGrade = self._poly_order
+            self._optimal_polynomial_grade = self._user_poly_order
 
-            print "Best fit polynomial order:"
-            print "\t: %d" % self.optimalPolGrade
+        # Attempting a parallel execution
 
+        channels = range(self._first_channel, self._n_channels + self._first_channel)
+        num_cpus = multiprocessing.cpu_count()
```
giacomov: The way of implementing parallel computation in 3ML is through the use of the Client from parallel/parallel_client.py . Look at the computation of contours in classicMLE/joint_likelihood.py for an example.

The reason to use that is that it allows a lot more flexibility through ipyparallel, for example allowing processes to be spawn across computer nodes. It also allows the user to cherry-pick what he wants to parallelize.

For example, if you run 100 jobs on a computer farm, you need to know in advance how many CPU each job will use, so the job should not decide by itself to use all the cpus of the node in which is running.

Please port this part to the Client interface, thanks!

giacomov: BTW, on the user side, a way to run the parallel computation of the contour is:

```python
with parallel_computation():
   jl.get_contours(...)
```

drJfunk:
@giacomov I was curious about this. I will make these changes now. They shouldn't conflict with anything you are doing. I also discovered that the way I was using parallel did not work on all GRBs. I was lucky with GRB 080916c because it had fewer events than newer GRBs. However, when the event list is too long, joblib crashed. 

I will use the client and do more testing to see if it works better. If not, it will just take some time to fit the background.

threeML/plugins/OGIP/eventlist.py
giacomov: I forgot to mention, this of course assume that there is a ipyparallel cluster running (with ipcluster). 

While it seems a bit cumbersome at the beginning, it actually allows a lot of flexibility. For example, I run a contour computation with 100k points from my a notebook on laptop using a machine of our farm at Stanford and had it in 10 minutes... :-)

PR #71, opened 2016-09-30
Adding chain consumer corner plot

drJfunk:
This adds the <optional> chain consumer plotter for corner plots. It adds a few more features such as the ability to overplot parameter distributions. 

The plots are also a bit more visually appealing. We may at some point just recode this library as it is not too complex. Anyhow, if the user does not have it, then it will warn them


threeML/bayesian/bayesian_analysis.py
```diff
@@ -584,6 +608,181 @@ def corner_plot(self, **kwargs):
 
             raise RuntimeError("You have to run the sampler first, using the sample() method")
 
+    def corner_plot_cc(self, sigmas=[0, 1, 2, 3], cloud=False, shade=True, shade_alpha=1., parameters=None,
```
giacomov: You should use the reStructuredText style for docstrings, as the rest of 3ML. Can you fix this and the other docstrings and request a merge again?

threeML/plugins/ogip.py
```diff
@@ -748,8 +750,24 @@ def _read_matrix(self, data, n_channels):
...
+                #
+                try:
+
+                    rsp[i, f_chan[i][j]: f_chan[i][j] + n_chan[i][j]] = matrix[i][m_start:m_start + n_chan[i][j]]
+
+                except(ValueError):
+
+                    rsp[i, f_chan[i][j] -1: f_chan[i][j]-1 + n_chan[i][j]] = matrix[i][m_start:m_start + n_chan[i][j]]
+
+
```
giacomov: I don't like this modification, as the response is read in a non-compliant way.

Do you have a matrix which was not read by the previous implementation. while it should have? If yes, please open an issue and provide the matrix, instead of modifying it here. 

This is a very critical part, we need to be extremely careful, so please revert to the original implementation and let's talk about this in the issue you are going to open.

threeML/bayesian/bayesian_analysis.py
drJfunk: Yup, pycharm was fooling with things.

threeML/plugins/ogip.py
drJfunk: I'm not sure how this got in here. I will fix

PR #72, opened 2016-10-03
fixed error in component plotting

drJfunk:
small one line change

### DIFF
--- a/threeML/io/model_plot.py
+++ b/threeML/io/model_plot.py
@@ -30,7 +30,7 @@ class SpectralPlotter(object):
 
         self._analysis_type = analysis._analysis_type
 
-        self.analysis = analysis
+        self._analysis = analysis
 
     def plot_model(self, x_unit='keV', y_unit='erg/(cm2 keV s)', sources_to_plot=[], summed=False, ene_min=10.,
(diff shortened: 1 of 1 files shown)


threeML/io/model_plot.py
```diff
@@ -618,9 +617,12 @@ def tmpflux(current_value):
 
                 first_derivatives.append(this_derivate)
 
+            # add the nuisance parameter first derivatives at the end (does this assumption always hold?)
```
giacomov: We cannot assume that first derivatives are always going to be the last parameters in the model, unfortunately.

Actually, I don't understand why you need to treat the nuisance parameters differently then the other parameters... the numerical differentiation should be able to compute zero derivatives without issues. This is also needed because there might be parameters in the model itself that have the same behavior, i.e., they give zero contribution to the variance at a particular energy. So we need to be able to handle this case.

Did you find out that that the numerical differentiation does not work for the nuisance parameters?

drJfunk:
The issue is that effective area corrections are not part of the photon model, so they do not appear in the equations. Therefore, I have to manually set the first derivatives. Is there something I'm missing?

PR #75, opened 2016-10-04
Combined changes

drJfunk:
Adds new corner plot

Adds some stats tools.

Moves ogip model plot to its own file for readability

modified ogip model plot to be able to use bayesian or male analysis.


threeML/plugins/FermiGBMTTELike.py
```diff
@@ -18,9 +18,9 @@
 __instrument_name = "Fermi GBM TTE (all detectors)"
 
 
-class FermiGBMLikeTTE(OGIPLike, PluginPrototype):
+class FermiGBMTTELike(OGIPLike, PluginPrototype):
```
giacomov: You don't need double inheritance anymore, since OGIPLike inherits already from PluginPrototype.

threeML/plugins/OGIP/eventlist.py
```diff
@@ -461,9 +461,12 @@ def peek(self):
...
             info_dict['Unbinned Fit'] = self._unbinned
 
-            S = li_and_ma(self._counts.sum(), self._poly_counts.sum())
+            sig =Significance(self._counts.sum(), self._poly_counts.sum())
 
-            info_dict['Li and Ma Sigma'] = S  # not sure if li and ma applies here
+            bkg_sig = np.sqrt( (self._poly_count_err**2).sum())
+
+
+            info_dict['Li and Ma Sigma'] = sig.li_and_ma_equivalent_for_gaussian_background(bkg_sig)  # not sure if li and ma applies here
```
giacomov: We shouldn't probably call this 'Li and Ma Sigma' since it is not ;-) Maybe "background-corrected significance"?

threeML/plugins/OGIP/likelihood_functions.py
```diff
@@ -114,4 +114,111 @@ def poisson_observed_gaussian_background(observed_counts, background_counts, bac
...
 
-    return np.sum(log_likes)
\ No newline at end of file
+    return np.sum(log_likes)
+
+
+# For some statistical functions (WAIC) it is desirable to look at the likelihood of each individual data point.
+# These functions return the vector of the observation for that purpose
+
+def poisson_log_likelihood_ideal_bkg_vector(observed_counts, expected_bkg_counts, expected_model_counts):
```
giacomov: ugh! :-)

Is this a copy and paste? I'd rather not do this, as possible errors in the original function will need to be corrected in two places...

Instead, let's have the old functions always return the vector of log likes, and the OGIPLike plugin perform a np.sum().

I can take care of this.

drJfunk: This must have been some leftover from very old things. I don't think I touched this.

threeML/plugins/OGIP/eventlist.py
drJfunk: Heh. I had the same thought. Vianello significance it is

PR #79, opened 2016-10-14
Pha write and temporal binner

drJfunk:
Introduces the ability to bin event lists and interfaces this through to TTELike. Also adds the nearly complete PHA writing class. I will still be adding this through the code as there are places to make this more useful. 

The PHA write still needs some adjustments for header variables but not for things that affect functionality. 

I'm pushing now as this fixes a bug in the eventList that causes import errors


threeML/plugins/OGIP/pha.py
```diff
@@ -485,3 +481,355 @@ def __repr__(self):
...
+                self._pseudo_time += pha_info[key].exposure
+
+                self._tstop[key].append(self._pseudo_time)
+
+    def _write_phaII(self, overwrite):
+
+        trigTime = None
+        clobber = False
+
+        # Assuming background and pha files have the same
```
giacomov: if I understand what this means, I think this should be enforced with an "assert" clause.

threeML/plugins/OGIP/response.py
```diff
@@ -135,6 +133,10 @@ def arf_filename(self):
 
         return self._arf_file
 
+    @property
+    def first_channel(self):
```
giacomov: can you add a short docstring? This property is exported to the public API

threeML/plugins/OGIP/response.py
```diff
@@ -151,9 +153,11 @@ def _read_matrix(self, data, header, column_name='MATRIX'):
             tlmin_fchan = header["TLMIN%i" % f_chan_column_pos]
 
         except(KeyError):
-            warnings.warn('No TLMIN keyword found. This DRM is improper. Assuming TLMIN=1')
+            warnings.warn('No TLMIN keyword found. This DRM is does not follow OGIP standards. Assuming TLMIN=1')
```
giacomov: typo

threeML/plugins/OGIPLike.py
```diff
@@ -56,7 +53,7 @@ def __init__(self, name, pha_file, bak_file=None, rsp_file=None, arf_file=None,
 
         # Get a PHA instance with the background
 
-        self._bak = self._get_pha_instance(bak_file, file_type='background')
+        self._bak = self._get_pha_instance(bak_file, file_type='background', spectrum_number=spectrum_number)
```
giacomov: originally the spectrum number was specified in the file name, as "my_spectrum.pha{1}". Is that still supported?

PR #103, opened 2016-11-18
added callback to goodness of fit minimizer setting

### LINKED ISSUE #102: GOF failing due to lacking callback
The minimizer in use member is now returning a callback and GOF was not expecting it.

drJfunk:
Fixes #102


giacomov:
Good catch!

PR #106, opened 2016-11-29
Config and colors for ogip

drJfunk:
Added configuration elements for OGIP and Fermi GBM plugins so that the user can modify certain colors and set their own defaults.


drJfunk:
I agree. Also should have a way to dynamically change things. I'll take a look!

____________________________________________________________________
J. Michael Burgess (http://www.jmichaelburgess.com/)
Oskar Klein Fellow (http://okc.albanova.se/)
Royal Institute of Technology (KTH)
Stockholm, Sweden

phone: +46 72.381.3020
Skype: james_m_burgess

PR #110, opened 2016-12-05
Config safety

grburgess:
This adds checks for the configuration file to make sure that the user does not introduce improper variables in their own configuration file. 

The intended behavior is as follows:

- try to load the local/user configuration file

- if file is corrupt, attempt to load the builtin/default configuration file.

Also added a py.test check that we have not corrupted the default config file

Added config variables for MLE contour and profile plotting.


drJfunk:
@giacomov, I'm failing to find where PyYAML supports preserving comments. Apparently, there is another package that does, but it is non-standard.

grburgess:
Another issue that I realized is that that there is not a clean way to make sure that all the keys exist in the user config without having them in some structure. Perhaps it is best to keep this the way it is?

giacomov:
I have merged this for the moment, it is not that important.

I know how to implement the thing about comments, but we have more important things to do right now. I'll open an issue as "enhancement" so we'll get to it sooner or later.

PR #115, opened 2016-12-15
Time dep likelihood

grburgess:
Passing to @giacomov


threeML/plugins/OGIPLike.py
```diff
@@ -906,6 +907,7 @@ def _get_observation_noise_model(self):
     observation_noise_model = property(_get_observation_noise_model, _set_observation_noise_model,
                                        doc="Sets/gets the noise model for the background spectrum")
 
+    @set_external_property
```
giacomov: We discussed this and you correctly found out that decorators are not inherited, so the way you implemented it is the only way using decorators.

However, I found another solution that would avoid having to decorate all plugins. I will implement it after the merge.

PR #124, opened 2016-12-19
Adds LLE plugin

grburgess:
This is a major rework of the EventList and associated classes. I've made a generic EventList and EventListLike to allow for the subtle difference between TTE and LLE. This is mostly involved with their handling of deadtime. The classes can be built upon as well. There is some repeat, but over all, it is very modular. 

@giacomov double check my livetime calculations please


threeML/plugins/FermiLATLLELike.py
```diff
@@ -0,0 +1,487 @@
...
+
+    @property
+    def livetime_start(self):
+        return self._tstart
+
+    @property
+    def livetime_stop(self):
+        return self._tstop
+
+    def _compute_mission_times(self):
```
giacomov: What is this for in this context? 

It is a nice functionality, maybe you could make it as a utility function under utils?

threeML/plugins/FermiLATLLELike.py
```diff
@@ -0,0 +1,487 @@
...
+            mission_df = pd.Series(mission_dict)
+
+            display(mission_df)
+
+        fermi_df = pd.Series(fermi_dict)
+
+        display(fermi_df)
+
+
+def lle_light_curve_plot(time_bins, cnts, bkg, width, selection, bkg_selections):
```
giacomov: At some point we will need to tackle this plotting thing, because having plotting routines all over the place is going to get messy. Maybe we should collect them all in one place. That way we can also reduce a lot the duplication of code.

threeML/plugins/OGIP/event_polynomial.py
```diff
@@ -191,39 +189,59 @@ def unbinned_polyfit(events, grade, t_start, t_stop, exposure, initial_amplitude
...
+            while (dof < 1 and len(initial_guess) > 1):
+                initial_guess = initial_guess[:-1]
+                polynomial = Polynomial(initial_guess)
+                log_likelihood = PolyUnbinnedLogLikelihood(events,
+                                                           polynomial,
+                                                           t_start,
+                                                           t_stop,
+                                                           exposure)
+
+        final_estimate = scipy.optimize.fmin(log_likelihood, initial_guess,
```
giacomov: I prefer to use the scipy.optimize.minimize function, so that you can change easily algorithm, instead of one of the more specialized versions like fmin.

threeML/plugins/OGIP/eventlist.py
```diff
@@ -1108,6 +1005,370 @@ def worker(start_index):
...
+        :return: exposure
+        """
+
+        # First see if the interval is completely contained inside a
+        # livetime interval. In this case we only compute that.
+
+
+
+
+        inside_idx = np.logical_and(self._live_time_starts < tmin, tmax < self._live_time_stops)
```
giacomov: shouldn't you use <=?

threeML/plugins/OGIP/eventlist.py
```diff
@@ -1108,6 +1005,370 @@ def worker(start_index):
...
+
+        # see if it contains elements
+
+        if self._live_time[inside_idx]:
```
giacomov: You need to consider the possibility of good time intervals and bad time intervals as well. The LLE file has a GTI extension that must be accounted for.

By construction, the GTI intervals must start and stop in correspondence of a FT2 entry, but you need to consider that as your light curve might cross a Bad Time Interval.

It could be easier to build first a function capable of giving the livetime between two arbitrary time intervals, and then simply apply it to all bins in the light curve.

Exploit the fact that data are ordered, so try using np.searchsorted if you need to search, instead of just (t >=tmin) & (t <=tmax), which need to cross all bins every time.

threeML/plugins/OGIP/eventlist.py
grburgess: I think this condition is covered below. I'll test again

grburgess:
Fixing up the branch

giacomov:
ok, but we could have merged this as is and keep it for future improvements. 

The code is good and correct, it just needs some more work to be fully general. Anyway, it's entirely up to you how you want to proceed.

thanks for doing this!

grburgess:
Ah! Well, let me get this one right. I'll fix it up this morning unless you want to merge now.

giacomov:
tomorrow is fine, it's late here and my wife would kill me if I work more ;-)

threeML/plugins/OGIP/eventlist.py
grburgess: Aha, I'm not doing searches, instead I'm building index filters... but if you want you have a better way let me know.

PR #129, opened 2016-12-22
Add info to sims

grburgess:
This adds parameter info to simulated data sets so that the user can see what they simulated.


threeML/plugins/OGIPLike.py
```diff
@@ -718,10 +721,35 @@ def get_simulated_dataset(self, new_name=None):
                 new_ogip_like._mask = original_mask
                 new_ogip_like._apply_mask_to_original_vectors()
 
+            # We want to store the simulated parameters so that the user
+            # can recall them
+
+            new_ogip_like._simulation_storage = copy.deepcopy(self._like_model)
```
giacomov: Astromodels provides the function clone_model for this. Can you use that instead of deepcopy?

grburgess:
Aha. Got it

threeML/plugins/OGIPLike.py
```diff
@@ -718,10 +722,35 @@ def get_simulated_dataset(self, new_name=None):
                 new_ogip_like._mask = original_mask
                 new_ogip_like._apply_mask_to_original_vectors()
 
+            # We want to store the simulated parameters so that the user
+            # can recall them
+
+            new_ogip_like._simulation_storage = clone_model(self._like_model)
+
+            # GV: It is probably safe to assume that there will be only one point source
```
giacomov: I do not understand this comment... where is this assumption made?

Good practice is to add an "assert" clause where this assumption is made, so that the user will be advised if the assertion is violated.

grburgess:
@giacomov I forgot to remove that comment. It was before I realized how to save the info. But now it does not matter!

PR #133, opened 2016-12-25
GBM/LLE catalog/ download 

grburgess:
Implements improvements to the GBM catalog such as different types of searches (parameter, trigger, etc)

Adds the ability to build 3ML models from the GBM catalog. 

Gets detector, background, and source intervals from the GBM catalog.

Adds GBM downloader of TTE and RSP files.

Adds LLE catalog with trigger and cone search. 

Adds LLE trigger download for event and RSP files.

Small clean up of some GBM functions.

new tests added


grburgess:
@giacomov I'm closing all three of these because the merge is going to be difficult. I'm going to make one big pull.

PR #134, opened 2016-12-31
Catalog additions and LLE

grburgess:
Adds the Fermi LLE plugin.

Adds (extends) the Fermi GBM and LLE catalogs and downloaders

Adds Swift GRB catalog

Adds simulation info to OGIP plugins

Tutorials added for LLE and catalogs.

Removes data from directories to save space now that the notebooks can download the data directly


threeML/catalogs/Fermi.py
```diff
@@ -1,58 +1,746 @@
...
+
+        ]
 
         new_table['ra'].format = '5.3f'
         new_table['dec'].format = '5.3f'
 
-        return new_table.group_by('Search_Offset')
+        return new_table.group_by('trigger_time')
+
+    def search_trigger_name(self, *trigger_names):
```
grburgess: Ok, this has changed except for a few things. I kept a general separate query for source names because it saves a lot of typing. A also left a specific query in Swift to save typing. Other than that, pandas handles everything.

threeML/catalogs/Fermi.py
```diff
@@ -1,58 +1,746 @@
...
+            best_fit_fluence = row['flnc_best_fitting_model'].split('_')[-1]
+
+            best_dict = {'fluence': best_fit_fluence, 'peak': best_fit_peak}
+
+            sources[name] = {'source'   : spectrum_dict, 'background': background_dict, 'trigger': trigger,
+                             'detectors': detector_selection, 'best fit model': best_dict}
+
+        return sources
+
+    def get_duration_information(self):
```
grburgess: Agreed, there is no simply a results property

threeML/catalogs/Fermi.py
```diff
@@ -267,3 +955,148 @@ def get_model(self, use_association_name=True):
...
+                          'trigger_time',
+                          'trigger_type'
+        ]
+
+        new_table['ra'].format = '5.3f'
+        new_table['dec'].format = '5.3f'
+
+        return new_table.group_by('trigger_time')
+
+    def search_trigger_name(self, *trigger_names):
```
grburgess: ;)

grburgess:
I agree with the comment about the downloaders. I've failed to find a more central solution at the moment. These will work for the foreseeable future, but when this is merged I will branch and brainstorm a bit more. The issue is that the current downloader needs to be wrapped. TTE files are huge and you do not want to download extras. I will get this though.

grburgess:
@giacomov finally builds thanks to your fix

threeML/plugins/Fermi_LAT/download_LLE_data.py
```diff
@@ -0,0 +1,278 @@
...
+
+
+
+
+
+
+    return download_info
+
+
+def _get_latest_version(filenames):
```
giacomov: Is this exactly the same as in the GBM one? If yes, after I merge, can you fix this by import this function directly from the GBM downloader, instead of copying/paste here?

threeML/plugins/FermipyLike.py
```diff
@@ -26,23 +26,22 @@
...
-               4: 'P8R2_TRANSIENT100_V6',
-               8: 'P8R2_TRANSIENT020E_V6',
-               16: 'P8R2_TRANSIENT020_V6',
-               32: 'P8R2_TRANSIENT010E_V6',
-               64: 'P8R2_TRANSIENT010_V6',
-               128: 'P8R2_SOURCE_V6',
-               256: 'P8R2_CLEAN_V6',
-               512: 'P8R2_ULTRACLEAN_V6',
-               1024: 'P8R2_ULTRACLEANVETO_V6',
+evclass_irf = {2    : 'P8R2_TRANSIENT100E_V6',
```
giacomov: ah ah! you have a strange taste... ;-)
Anyway, please refrain from doing all these small changes in pull requests because it makes the number of files I need to check very large... I authorize you to make aesthetic changes like these directly in the trunk :-)

PR #136, opened 2017-01-05
Rsp2 and bug fix

grburgess:
A few cleanups in EventList and RSP2 handling


threeML/plugins/EventListLike.py
```diff
@@ -73,11 +71,55 @@ def __init__(self, name, event_list, background_selections, source_intervals, rs
...
+        try:
+
+            OGIPLike.__init__(self,
+                              name,
+                              pha_file=self._observed_pha,
+                              bak_file=self._bkg_pha,
+                              rsp_file=rsp_file,
+                              verbose=verbose)
+
+        except(NegativeBackground):
```
giacomov: If I understand well what is going on here, I am not sure this is right.

A negative background does not necessarily mean that the polynomial grade is too high... I'd rather throw an exception and advise to change the background off-pulse intervals.

threeML/plugins/OGIP/eventlist.py
```diff
@@ -1062,12 +1113,10 @@ def __init__(self, arrival_times, energies, n_channels, start_time=None, stop_ti
 
             self._dead_time = None
 
-
-
-    def _exposure_over_interval(self, tmin, tmax):
+    def exposure_over_interval(self, tmin, tmax):
```
giacomov: Is this method supposed to be called by the user?

If yes, then you should add a detailed docstring.

If not, then you should call it _exposure_over_interval (also in the parent class).

threeML/plugins/EventListLike.py
```diff
@@ -1,10 +1,8 @@
 __author__ = 'grburgess'
 
 import numpy as np
-import matplotlib.pyplot as plt
-import pandas as pd
-import warnings
-import re
+
+from threeML.exceptions.custom_exceptions import custom_warnings, NegativeBackground
```
grburgess: Aha, ok. I was trying to keep them central. I'll stop that!

threeML/plugins/EventListLike.py
grburgess: Fixed.

threeML/plugins/OGIP/eventlist.py
grburgess: It is not supposed to be called by the user, but it is called from an EventList object externally. Therefore, I believe that it should be public. (unless I misinterpret the public/private concept in python) I have added a detailed doctstring.

Note, EventList is not exposed to the user.

PR #140, opened 2017-01-12
Improve model plotting

grburgess:
Sorry for the crazy number of files changed. It is mostly data and from my dumb idea to have the data downloaded. 

This adds a new system for flux calculations and model plotting that is modular and easily extendable to extended sources. 

For now, only point sources have dedicated classes. The calculations are fast now that I have delayed using units until the end of the process. 

flux and plotting of point sources is now done with functions instead of classes. Key words remain the same. examine the updated demos.

Also, we now have better and full support of using energy/wavelength/frequency for the energy axis of plotting or flux. This is done using native astropy tools.

I also fixed a few bugs. The downloader was ignoring compressed files and re downloading them... but no longer. 

I reorganized the io folder as it was getting big.


grburgess:
PS: coverage decreased... but I wanted to fix that on my own after you review.

grburgess:
@giacomov I have successfully updated this to the new framework. 

I kept the same idea of have "handlers" for fitted sources except much more of the work is being done by RandomVariates and the implementation is MUCH cleaner now that error propagation is uniform between MLE and Bayesian. This also got rid of the nasty multiple inheritance.

One issue was the fact that you can not use astropy units  with or make arrays of RandomVariates because the constructor becomes confused and turns the RV into a regular array. I looked at modifying the constructor, but it would have been really ugly and the VariatesContainer I have made is a much simpler solution that also deals with unit conversion easily. 

The frame work is extendable to polarization plotting, extended objects, etc. with dedicated subclasses. 

The actual plot and flux functions are a bit verbose, but this is to handle all the cases that may come up. I've tried to keep duplicate code to a minimum except where it is needed.

grburgess:
@giacomov also fixed a bug in analysis results. you were calling equal_tail for both equal_tail and hpd

PR #148, opened 2017-01-21
From event list

grburgess:
this integrates the new PHAII class throughout the code and gets rid of the PHAContainer class. Some slight modifications to the the fits_file class to accommodate the changes.

@giacomov, I updated the VERITASLike. Please ensure that this works as I cannot test it.


grburgess:
@giacomov I think a lot of the functionality will come and and be smoother when we move to spectrum.

PR #151, opened 2017-01-23
Integrate time_interval class

grburgess:
the time interval class its a nice addition. I've implemented it into the event list dependent classes which greatly simplified some code and also made the code much more readable. 

Additionally, I extended the classes a bit for things I needed and perhaps some future functionality. 

I recommend merging this to the master only after my other commits have been put it such as #148 .
I can also wait on that and merge myself.  I also integrated #150 here otherwise it would not work.


grburgess:
@giacomov the LAT download is failing repeatedly. Otherwise all checks work. still, it screws up the coverage without the test

grburgess:
@giacomov all has pass now. We still need to watch the LAT download test.

PR #156, opened 2017-01-26
Add repr to plugins

grburgess:
moved from using peek to using __repr__
 additionally added some convenience properties to OGIP


grburgess:
@giacomov as long as the tests pass, I've made the .display() feature you have requested

PR #158, opened 2017-01-30
switched temporal binner to a child of time interval set

grburgess:
This just makes for some cleaner code and switches the temporal bins in the event list to being held in time intervals.

Once #156 is merged, this will have some powerful repr abilities


giacomov:
Nice!

You might want to consider renaming the methods from_* to just bin_by_significance or similar. I understand why you used the from_* syntax, but it might end up being a little confusing for the user, who does not need to know about the class structure.

PR #163, opened 2017-02-12
More spectrum work

grburgess:
This actually moves the Spectrum class into the working part of the code. OGIPLike has been moved to a subclass of SpectrumLike. There was a lot of code cleanup to make things smooth and a lot of duplicated code was centralized.


grburgess:
@giacomov a couple of changes that may alter your current code:

display_ogip_model_counts -> display_spectrum_model_counts

OGIPLike(pha_file, bak_file) -> OGIPLike(observation, background)

threeML/utils/histogram.py
```diff
@@ -0,0 +1,267 @@
...
+            new_sys_errors = other.sys_errors
+
+        else:
+
+            new_sys_errors = None
+
+        new_contents = self.contents + other.contents
+
+
+        # because Hist gets inherited very deeply, when we add we will not know exactly
```
giacomov: You should call the constructor, instead of using copy, like:

new_hist = type(self)(..., new_contents, new_errors, new_sys_errors)

This of course works if the child class keeps the same calling sequence for the constructor.

Probably a better way is to rename this method _add_engine (or whatever) and have it return new_contents, new_errors, new_sys_errors.

Then you define a __add__ method like this:

```python
def __add__(self, other):
    c, e, se = self._add_engine(other)
    intervals = ...
    return type(self)(intervals, c, e, se)
```

If a child class changed the syntax of the constructor, it will need to implement a new __add__ method which will call _add_engine() and then build a new instance with the results.

grburgess: Aha! Sorry, these comments didn't pop up in my RSS read... so I missed them. I'll make these changes.

threeML/plugins/OGIPLike.py
```diff
@@ -1,1420 +1,72 @@
...
+        if background is None:
 
-        :param value: new value (default: 1, i.e., no correction)
-        :return:
-        """
+            background = pha.background_file
 
-        self._nuisance_parameter.value = value
-        self._nuisance_parameter.fix = True
+            assert background is not None, "No background file provided, and the PHA file does not specify one."
```
grburgess: :) I was curious about that this morning and saw this just now. I can add this right now.

threeML/plugins/SpectrumLike.py
```diff
@@ -0,0 +1,1495 @@
...
+            # can recall them later
+
+            new_spectrum_plugin._simulation_storage = clone_model(self._like_model)
+
+            # TODO: nuisance parameters
+
+            return new_spectrum_plugin
+
+    @classmethod
+    def _new(cls,*args,**kwargs):
```
grburgess: They were a way for me to implement a class method inside an instance method. But now I see you type(self)(*args) idea and will switch to that.

PR #165, opened 2017-02-15
Added `set_bin_list` method to `HAWCLike` to allow use of non-contiguous …

sammarinelli:
…sets of bins and bins with arbitrary string names now that these features are supported by LiFF.  Also added plotting of pulls as alternative to residuals in `HAWCLike.display`, turned on using `pulls=True`.

### DIFF
--- a/threeML/plugins/HAWCLike.py
+++ b/threeML/plugins/HAWCLike.py
@@ -5,6 +5,8 @@ import sys
 import matplotlib.pyplot as plt
 import numpy as np
 from astromodels import Parameter
+from cthreeML.pyModelInterfaceCache import pyToCppModelInterfaceCache
+from hawc import liff_3ML
 from matplotlib import gridspec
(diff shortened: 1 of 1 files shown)


giacomov:
@sammarinelli thanks for this! 

The tests are failing, but I suspect it is because the docker container we are using for the test contains a version of LiFF which is not updated enough.

Which version of LiFF should I install to support this feature?

sammarinelli:
I talked with @giacomov offline, but I'll post our conclusion here for anyone else following along.  The new functionality requires a more recent version of AERIE than the latest release, which is from last April.  @giacomov is going to try building against the AERIE trunk instead.

giacomov:
@sammarinelli I updated the test environment with the most recent version of aerie, and now all tests are passing successfully.

Thanks for this contribution, I am going to merge it now.

PR #168, opened 2017-02-21
added parallel abilites to error propagation

grburgess:
Added the parallel client to the error propagation to speed it up


threeML/utils/fitted_objects/fitted_source_handler.py
```diff
@@ -143,14 +155,143 @@ def _evaluate(self):
             # scroll through the independent variables
             n_iterations = np.product(self._out_shape)
 
-            with progress_bar(n_iterations, title="Propagating errors") as p:
+            if not threeML_config['parallel']['use-parallel']:
```
giacomov: This looks overcomplicated to me.

Have you looked at the code of ParallelClient? There are functions to easily perform parallel computation by just providing the worker (they even make the progress bar for you). The code you copied from (from the contour procedures) is old, and it cannot be updated because of the inherent way a contour plot is computed.

grburgess:
Ah ok. I recently understood the guts if ipyparellel, so I'll take another attempt at this.

grburgess:
@giacomov I see your issue with the reused code, but I forgot that there was an issue with the fact that I need to reorder the grid if it is changed by the number of engines. Therefore, some of this code needs to stay.

PR #171, opened 2017-03-22
Photo like

grburgess:
This PR implements the basic photometry plugin. Filters from SVO are added.

Some small refactoring of plotting routines to avoid duplicated code. 

I also modified slightly XYLike simulation to allow for different constructor args.

YES... a lot of files but these are the filter curves!


giacomov:
Awesome!

I'm merging into the master and review it offline from there, it is impossible for me to review this on github, with these many files.

PR #174, opened 2017-04-03
Fix bayes plot

grburgess:
added functionality to plotting and docs for plotting


grburgess:
For some reason codecov did not issue a report because of an idiot move I made. however, I visually inspected the coverage and it is fine. This is ok to merge.

I just cleaned up some bugs in plotting. No big changes.

Issue #175, opened 2017-04-04
Countors/ corner plots from analysis results


grburgess:
@giacomov the ability to save analysis results is really great!

However, it would nice to be able to post compute some quantities from them. For example, with a joint likelihood set, it is not currently possible to make contour plots, but if we connected them to the AnalysisResults class, then we could post compute these.

 Of course, this requires data which is not stored, so this may in fact be impossible... but, we do have samples from the covariance, and actual samples in the Bayesian case. Therefore, maybe we connect the corner plots to analysis results directly? You would do like this:

```
jl.results.corner_plot()
bayes.results.corner_plot()

```

where the chains are the samples stored in the AR.

this is of course different than profiling the likelihood in the MLE case, but one at least can have a look at the distributions AFTER an analysis has been saved. In any case, being able to look at the Bayesian corner plots without having to keep them all alive in memory would be hugely beneficial.

PR #176, opened 2017-04-04
Move corner to ar

grburgess:
This moves the corner plots to the analysis results in re: to #175 . It actually makes some nice plots for the MLE fits as well! 

Now one can make corner plots from the saved analysis results directly. I still left an interface directly from the BayesianAnalysis class to keep old functionality.


grburgess:
@giacomov I am not sure what is up with one of the coverage tests, but the other is correct.

giacomov:
This is very convenient, but I am a little worried because it might be confusing.

Making a "corner plot" with samples from the covariance matrix is a bit of a stretch. You might as well just plot a multi-dimensional Gaussian with the covariance matrix and get the same result.

It is not the same as plotting a contour plot with jl.get_contour, because the latter does not assume linear correlation among the parameters (and hence can produce contours which are not elliptical).

But of course, as you note, jl.get_contour works only with the plugins loaded because it needs to refit for each point in the grid of the contour.

So, maybe it is worth pursuing this, but I would add a warning that gets printed every time you make a contour plot from the analysis results, saying something like: "Careful: these plots are an approximation, as they assume linear correlation among parameters. To account for non-linear correlation use the get_contour method of the JointLikelihood object."

grburgess:
My thoughts exactly. I considered putting a warning sign. I can do that. 

I suppose there is not a very simple way to save a JL fit, but they are much faster than a Bayesian fit. The question is: what to do for a JL jet?

grburgess:
@giacomov warning added in last push

giacomov:
JointLikelihoodSet has a .results property returning the set of results, and a .write_to method which writes one result set for each model.

The produced FITS file can be read back with load_analysis_results.

The AnalysisResultsSet object has also methods to associate a label and some values to the various results. Say you are performing a time-resolved analysis, and you have a set of 10 results. You can use set_bins or set_x of the AnalysisResultsSet object to associate the start and stop of the bins (or the center of the bin) to the results, before writing them to the file. This makes plotting from the analysis results set easier.

grburgess:
@giacomov, I'm not sure I follow. what I meant was plotting contours from a JointLikelihoodSet.

Another option to a warning is to only have the corner plot associated with the BayesianResults object. What do you think is best?

giacomov:
once you have the AnalysisResultsSet object, you can get an AnalysisResults object as:
```
analysis_results_first_interval = analysis_results_set[0]
```
and use it to get the contour plot.

Or you mean getting the contours from a joint likelihood object for each interval? In that case, you need to use the JointLikelihoodSetAnalyzer class, get the best fit model and the data using the restore_best_fit method, and then you can recreate the JointLikelihood object from which you can get the contours.

grburgess:
Nice. That was what I was curious about. And now I see that you can easily use this with saved JL results as well. So, shall I simply move the corner plot to the BayesianResults?

giacomov:
yes, that's a good idea.

PR #183, opened 2017-04-26
Bayes model selection

grburgess:
This adds additional information to the analysis results and gets rid of the ModelComparison class. Now, AIC, DIC, BIC are calculated after a fits and saved in the AR. This requires an additional universal abc method in plugin prototype to return the number of data points used. For binned spectra, this is just the total number of channels used in the fits. For unbanned spectra, it is the number of points used to evaluate the likelihood. @giacomov I have put some dummy functions in for the Fermi/HAWC plugins for you to update. All other plugins should be taken care of, except that I think you switched VERITAS to its own plugin.


grburgess:
Failing dude to the merge with the master branch which was failing. I will try to fix.

threeML/classicMLE/joint_likelihood.py
```diff
@@ -283,6 +284,8 @@ def fit(self, quiet=False, compute_covariance=True):
 
         total = 0
 
+        total_number_of_data_points = 0
```
giacomov: You should add meaningful comments throughout the code you are committing to explain what is happening and why, so we will remember when we'll change the code years from now...

Normally it should be one line / one comment, or close to that ;-)

threeML/plugin_prototype.py
```diff
@@ -125,3 +125,14 @@ def inner_fit(self):
         logLike value.
         """
         pass
+
+    @abc.abstractmethod
```
giacomov: Please do not make this an abstract method. Instead, provide here an implementation that issue a warning like "get_number_of_data_points not implemented, values for statistical measurements such as AIC or BIC are unreliable", and then returns 1. 

This way you can remove the placeholders in all plugins, and we will remember  that we need to fix this eventually.

threeML/classicMLE/joint_likelihood.py
grburgess: ah, yes. will do. Good optimistic thinking.

threeML/plugin_prototype.py
grburgess: Cool. The only issue is that it will crash the AnalysisResults unless a number is returned.

grburgess: Ah.... sorry I missed part of your comment 👎

PR #184, opened 2017-04-26
allow for force write of rsps

grburgess:
This allows for forcing the write of an RSP even when it is not required. Typically a write only happens when we run across and RSP set. Now you can force a write to make the DRM in 3ML (so proper OGIP) format. Clears up some issues with GBM reps.

### DIFF
--- a/threeML/plugins/DispersionSpectrumLike.py
+++ b/threeML/plugins/DispersionSpectrumLike.py
@@ -104,7 +104,7 @@ class DispersionSpectrumLike(SpectrumLike):
 
         return super_out.append(the_df)
 
-    def write_pha(self, filename, overwrite=False):
+    def write_pha(self, filename, overwrite=False, force_rsp_write=False):
         """
 
         :param filename:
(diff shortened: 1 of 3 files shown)


threeML/plugins/OGIP/pha.py
```diff
@@ -69,6 +69,8 @@ def write(self, outfile_name, overwrite=True):
         :return:
         """
 
+        self._force_rsp_write = force_rsp_write
```
giacomov: It would be better to simply pass force_rsp_write as a parameter to self._append_ogip.

In general, no attribute of the class should be set for the first time outside the constructor (or if really necessary in methods called by the constructor).

The reason is that sometimes you need to know the state of the class (for example to serialize it with pickle or dill for parallel processing), and a state that changes depending on the interaction with the user is going to be very difficult to handle.

threeML/plugins/DispersionSpectrumLike.py
```diff
@@ -104,7 +104,7 @@ def _output(self):
 
         return super_out.append(the_df)
 
-    def write_pha(self, filename, overwrite=False):
+    def write_pha(self, filename, overwrite=False, force_rsp_write=False):
```
giacomov: Can you update the docstrings of the methods accordingly?

grburgess: fixed

PR #188, opened 2017-04-30
Fits file formats

### LINKED ISSUE #187: reading of certain fits types failing.
I found a bug in our fits reader that we need to correct. 

While the GBM CSPEC files claim to contain integer counts of type 128I -> signed short,
they are actually 128U -> unsigned short. This is fine and numpy dtype is mapped to:

```
_NUMPY_TO_FITS_CODE = {

    # Integers

    np.int16: 'I', # what the file claims we have in the header
    np.int32: 'J',
    np.int64: 'K',
    np.uint16: 'U', # what we have
    np.uint32: 'V',

    # Floating point

    np.float32: 'E',
    np.float64: 'D',

    }

```

however, astropy does not support this data type:

```
# mapping from TFORM data type to numpy data type (code)
# L: Logical (Boolean)
# B: Unsigned Byte
# I: 16-bit Integer
# J: 32-bit Integer
# K: 64-bit Integer
# E: Single-precision Floating Point
# D: Double-precision Floating Point
# C: Single-precision Complex
# M: Double-precision Complex
# A: Character
FITS2NUMPY = {'L': 'i1', 'B': 'u1', 'I': 'i2', 'J': 'i4', 'K': 'i8', 'E': 'f4',
              'D': 'f8', 'C': 'c8', 'M': 'c16', 'A': 'a'}

# the inverse dictionary of the above
NUMPY2FITS = {val: key for key, val in iteritems(FITS2NUMPY)}
# Normally booleans are represented as ints in pyfits, but if passed in a numpy
# boolean array, that should be supported
NUMPY2FITS['b1'] = 'L'
# Add unsigned types, which will be stored as signed ints with a TZERO card.
NUMPY2FITS['u2'] = 'I'
NUMPY2FITS['u4'] = 'J'
NUMPY2FITS['u8'] = 'K'
# Add half precision floating point numbers which will be up-converted to
# single precision.
NUMPY2FITS['f2'] = 'E'

# This is the order in which values are converted to FITS types
# Note that only double precision floating point/complex are supported
FORMATORDER = ['L', 'B', 'I', 'J', 'K', 'D', 'M', 'A']

# Convert single precision floating point/complex to double precision.
FITSUPCONVERTERS = {'E': 'D', 'C': 'M'}

# mapping from ASCII table TFORM data type to numpy data type
# A: Character
# I: Integer (32-bit)
# J: Integer (64-bit; non-standard)
# F: Float (64-bit; fixed decimal notation)
# E: Float (64-bit; exponential notation)
# D: Float (64-bit; exponential notation, always 64-bit by convention)
ASCII2NUMPY = {'A': 'a', 'I': 'i4', 'J': 'i8', 'F': 'f8', 'E': 'f8', 'D': 'f8'}

# Maps FITS ASCII column format codes to the appropriate Python string
# formatting codes for that type.
ASCII2STR = {'A': '', 'I': 'd', 'J': 'd', 'F': 'f', 'E': 'E', 'D': 'E'}

# For each ASCII table format code, provides a default width (and decimal
# precision) for when one isn't given explicitly in the column format
ASCII_DEFAULT_WIDTHS = {'A': (1, 0), 'I': (10, 0), 'J': (15, 0),
                        'E': (15, 7), 'F': (16, 7), 'D': (25, 17)}

```

So I suggest we remap the unreadable ints to a different format. I am not sure what type of effect this will have.

grburgess:
Fixes #187 and adds tests to make sure what is input and written to disk does not change.

### DIFF
--- a/threeML/test/test_fits_file.py
+++ b/threeML/test/test_fits_file.py
@@ -1,0 +1,68 @@
+
+from threeML.io.fits_file import FITSExtension, FITSFile
+import numpy as np
+import astropy.io.fits as fits
+
+import pytest
+
+
+class DUMMYEXT(FITSExtension):
+
+    def __init__(self, test_value):
+
+        data_list = [('TEST_VALUE', test_value)]
+
+        super(DUMMYEXT, self).__init__(tuple(data_list), (('EXTNAME', 'TEST', 'Extension name'),) )
+
+
+class DUMMYFITS(FITSFile):
+
+    def __init__(self, test_value):
+
+        dummy_extension = DUMMYEXT(test_value)
+
+        super(DUMMYFITS, self).__init__(fits_extensions=[dummy_extension])
+
+
+def test_fits_file():
+
+    dtypes = [np.int16,np.int32,np.int64,np.uint16,np.uint32,np.float32,np.float64]
+    dtype_keys = ['I','J','K','I','J','E','D']
+
+
+    for i, dt in enumerate(dtypes):
+
+        test_values = np.ones(10,dtype=dt)
+
+        dummy_fits = DUMMYFITS(test_value=test_values)
+
+        assert len(dummy_fits._hdu_list) == 2
+
+        assert dummy_fits.index_of('TEST') == 1
+
+        assert dummy_fits['TEST'].header['TFORM1'] == dtype_keys[i]
+
+        assert len(dummy_fits['TEST'].data['TEST_VALUE'] == test_values)
+
+
+        file_name = 'test_fits%d.fits'%i
+
+        dummy_fits.writeto(file_name,overwrite=True)
+
+        with pytest.raises(IOError):
+
+            dummy_fits.writeto(file_name, overwrite=False)
+
+
+        read_dummy_fits = fits.open(file_name)
+
+
+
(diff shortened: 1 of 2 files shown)


threeML/test/test_fits_file.py
```diff
@@ -0,0 +1,68 @@
...
+
+        dummy_fits = DUMMYFITS(test_value=test_values)
+
+        assert len(dummy_fits._hdu_list) == 2
+
+        assert dummy_fits.index_of('TEST') == 1
+
+        assert dummy_fits['TEST'].header['TFORM1'] == dtype_keys[i]
+
+        assert len(dummy_fits['TEST'].data['TEST_VALUE'] == test_values)
```
giacomov: what does this test?

If I understand, the part "dummy_fits['TEST'].data['TEST_VALUE'] == test_values" will give you a boolean vector with length 10, so len(...) will give you 10 whether all the elements in the boolean vector are True or not. Am I wrong?

You could use np.alltrue() instead, which will return False if and only if at least one element of the boolean vector is False.

grburgess: doh. I always forget this. You are correct. Actually, I even had a small bug here with the len(...) argument. I'll fix it up.

PR #190, opened 2017-05-14
moved to custom AB getter

grburgess:
The speclite library was not doing the units conversions carefully which could result in magnitudes being improperly normalized. However, they utilities they use for these conversions are exposed and I have been able to do the calculations myself.


grburgess:
Ok, I fixed up the test. Now the optical photometry is giving proper results. I will stick with speclite for the moment.

PR #192, opened 2017-05-16
Integral-friendly

volodymyrss:
* fixing issue in __repr__ of  with spectra with no background
* more permissive OGIP reading, compatible with INTEGRAL spectra


threeML/plugins/OGIP/pha.py
```diff
@@ -543,8 +544,25 @@ def from_fits_file(cls,fits_file):
...
 
 
-            spectrum = FITSExtension.from_fits_file_extension(f['SPECTRUM'])
-
+            if 'SPECTRUM' in f: 
+                spectrum_extension=f['SPECTRUM']
+            else:
+                warnings.warn("unable to find SPECTRUM extension: not OGIP PHA!")
+                spectrum_extension=None
+                for extension in f:
```
giacomov: you can get more easily the value of a keyword using the .get() method of the header:

http://docs.astropy.org/en/stable/io/fits/api/headers.html

It returns the value if present, or None if not present. So for example:

```
if 'HDUCLASS' in extension.header and extension.header['HDUCLASS'] == 'OGIP'...
```
becomes:
```
if extension.header.get('HDUCLASS')=='OGIP'...
```
You can probably restructure this as:
```
for extension in f:
    hduclass = extension.header.get("HDUCLASS")
    hduclass1 = extension.header.get("HDUCLAS1")
    
    if hduclass == 'OGIP' and hduclas1 == 'SPECTRUM':
        # Found spectrum extension
        spectrum_extension = extension
        # Issue warning
        warnings.warn("File has no SPECTRUM extension, but found a spectrum in extension %s" % (spectrum_extension.header.get("EXTNAME")))
        # Reset name to SPECTRUM
        spectrum_extension.header['EXTNAME'] = 'SPECTRUM'
        # No need to continue looping
        break
```

(I didn't test the above code, there might be some small problem)

threeML/plugins/OGIPLike.py
```diff
@@ -118,7 +118,10 @@ def _output(self):
 
         superout = super(OGIPLike, self)._output()
 
-        this_out = {'pha file': self._observed_spectrum.filename, 'bak file': self._background_spectrum.filename}
+        this_out = {
+                    'pha file': self._observed_spectrum.filename,
+                    'bak file': self._background_spectrum.filename if self._background_spectrum is not None else None
+                    }
```
grburgess: Could you change the nested if statement to such that it reads like this:
```python

 if self._background_spectrum is not None:
    bak_file =  self._background_spectrum.filename

else:
    bak_file = None


this_out = {
                     'pha file': self._observed_spectrum.filename,
                     'bak file': bak_file

```
It is a bit easier to read. Ideally, I should make a dummy background spectrum in the future. Perhaps make similar changes throughout. Thanks for catching this!

threeML/plugins/SpectrumLike.py
```diff
@@ -66,10 +66,10 @@ def __init__(self, name, observation, background, verbose=True):
         self._observed_spectrum = observation  # type: BinnedSpectrum
 
         self._observed_counts = self._observed_spectrum.counts  # type: np.ndarray
-
+            
         if background is None:
 
-            self._background_spectrum = background  # type: BinnedSpectrum
+            self._background_spectrum =  None
```
grburgess: Good catch. This is much clearer.

volodymyrss:
Thanks!

I have updated accordingly.
I did not make any attempt to accurately compute significance in no-background case.

I am trying to use it and it seems to work well for my goals, albeit some small adjustments for INTEGRAL here and there.
I am glad I can contribute a little be to such an interesting project!

grburgess:
Thanks @volodymyrss! Also, if you are having install issues please pot an issue so that we can track and have an open conversation where other users can follow.

PR #196, opened 2017-05-17
Support for adding instances of BinnedSpectrum

volodymyrss:
I really needed adding of the spectra, in the simplest case, to accumulate exposure. 
I need two methods, average rate, and weighted average.

There were also some small changes here and there that were required.

I am not not too sure about the returning None errors in Poisson spectra, instead of assert. I just tried to minimize multiple "if self.is_poisson: x=None" statements.

### DIFF
--- a/threeML/data_list.py
+++ b/threeML/data_list.py
@@ -21,24 +21,24 @@ class DataList(object):
 
         for d in data_sets:
 
-            if d.get_name() in self._inner_dictionary.keys():
+            if d.name in self._inner_dictionary.keys():
 
-                raise RuntimeError("You have to use unique names for data sets. %s already exists." % (d.get_name()))
+                raise RuntimeError("You have to use unique names for data sets. %s already exists." % (d.name))
 
             else:
 
-                self._inner_dictionary[d.get_name()] = d
+                self._inner_dictionary[d.name] = d
 
     def insert(self, dataset):
 
         # Enforce the unique name
-        if dataset.get_name() in self.keys():
+        if dataset.name in self.keys():
 
             raise RuntimeError("You have to use unique names for data sets. %s already exists." % key)
 
         else:
 
-            self._inner_dictionary[dataset.get_name()] = dataset
+            self._inner_dictionary[dataset.name] = dataset
 
     def __getitem__(self, key):
 
--- a/threeML/plugins/FermiLATLike.py
+++ b/threeML/plugins/FermiLATLike.py
@@ -300,8 +300,6 @@ class FermiLATLike(PluginPrototype):
         '''
         return self.name
 
-    pass
-
     def setInnerMinimization(self, s):
 
         self.innerMinimization = bool(s)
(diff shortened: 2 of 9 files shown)


giacomov:
Thanks!

However, what is the use case for this?

What do you need to achieve that cannot be achieved by simply loading one plugin instance per spectrum, and using them in a JointLikelihood object? Something like this:

```python
sp1 = SpectrumLike(...)
sp2 = SpectrumLike(...)
data = DataList(sp1, sp2)

model = ...

jl = JointLikelihood(model, data)
```
I know that summing spectra has been a widely-used procedure in the X-ray community for years, but I don't like it: it is very error-prone because the handling of the background and especially of the response is tricky (the proper way of summing the response properly varies case by case). 

If you only need to put together different spectra to gain statistic, using them in a joint likelihood achieves that already while maintaining each spectrum with its own background and response.

volodymyrss:
Indeed, I tried to use joint likelihood when possible.
I understand that this is the good way.
Summing kills information.

However, for us it's common to have up to ~100.000 spectra for a single source. I tried with a 100 and it's not quite as interactive, though still ok. 

For response I put so far an assert that it is identical.
The spectra are already background(and other \source)-subtracted. 
The summing could be limited to simple cases like this. 
Potentially we could do the total joint likelihood fit of course, together with background, this would make 100.000 spectra and  200 x 100.000 sources (different parameters in each observation). Would it work with 20M sources?
Even if it will, it will take time, each attempt.

For exploration and a first look it is practical to see the merged spectrum.
Finally and related, it's nice for plotting purposes (though this could be achieved differently I guess)

volodymyrss:
It's an interesting question of course. I do not know a  better way right now than to stack spectra at least partially. A way which would be also be quite practical, for the reasons I listed.
Though really I made a pull request because J Michael mentioned it was in plans anyway.

giacomov:
ok, this is indeed a good use case for something like this. Yes, while astromodels is very lightweight, I never had in mind a use case with 20 million sources! So it would probably become unbearably slow.

I think having the assertion that the response is identical is the way to go, so we will not have in the future people shooting themselves in the foot by summing spectra which are not supposed to be summed.

So let's go ahead with this. I see that the tests are failing, but probably you are still working on it?

volodymyrss:
I left the tests that verify that accessing BinnedSpectrum.count_errors (and similar) for poisson spectra should raise an exception. This is related to the somewhat higher impact change I made (mentioned in the pull request).

replacing 
```python
assert self.is_poisson == False, "Cannot request errors on rates for a Poisson spectrum"
return self._errors * self.exposure
```
with
```python
if self.is_poisson:
    return None
else:
    return self._errors * self.exposure
```

The exception would come up often when cloning or adding, leading to many repetitions of something like this:

```python
if not self.is_poisson:
    function(...,self.count_errors)
else:
    function_alternative(...)
```
at many more technical stages there is no need to discriminate between poisson and gauss cases, so code repetition can be avoided by assuming that errors will be `None` in the poisson case.  


If you prefer, I can add all those if's, not a big issue.

PR #197, opened 2017-05-18
Convert to Python3

ralic:
using : find . -name '*.py' | xargs 2to3 -w


giacomov:
Hi,
while we wrote the code thinking about this, it is not yet time to convert to py3. Some of the instruments we are using 3ML with (for example Fermi) have a software which is not compatible with python3.

Do you have a specific need for this?

ralic:
I may try to work on Fermi packages. I checked the website on Nasa. Is the source code from this page ?
https://fermi.gsfc.nasa.gov/ssc/data/analysis/software/

ralic:
Or here ?
https://fermi.gsfc.nasa.gov/ssc/data/analysis/scitools/overview.html

PR #198, opened 2017-05-18
plotting ratio residuals

volodymyrss:
Additionally to sigma residuals it can be useful sometimes to see ratio of the residual to the model.

Sometimes because of complex dependency of detector efficiency on energy, the uncertainties change strongly and non-monotonically with energy. 
If there is a strong monotonic residual, it will appear to have features. The features correspond to sensitivity to the model, but they may be misinterpreted as line features.
Residual ratio does not have this problem (it has others)

### DIFF
--- a/threeML/io/plotting/data_residual_plot.py
+++ b/threeML/io/plotting/data_residual_plot.py
@@ -15,15 +15,17 @@ class ResidualPlot(object):
         :param kwargs:
         """
 
-        self.ratio_residuals = False
+        self._ratio_residuals = False
         if 'ratio_residuals' in kwargs:
-            self.ratio_residuals = bool(kwargs.pop('ratio_residuals'))
+            self._ratio_residuals = bool(kwargs.pop('ratio_residuals'))
 
         self._fig, (self._ax, self._ax1) = plt.subplots(2, 1, sharex=True,
                                                         gridspec_kw={'height_ratios': [2, 1]}, **kwargs)
 
 
-
+    @property
+    def ratio_residuals(self):
+        return self._ratio_residuals
 
 
     def add_model_step(self, xmin, xmax, xwidth, y, label, color='r'):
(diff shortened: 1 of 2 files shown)


giacomov:
Thanks for this addition!

Can you post here an example on how to use it?

volodymyrss:
Sure, here it is:

```python
threeML.display_spectrum_model_counts(jl,ratio_residuals=True)
```

threeML/io/plotting/data_residual_plot.py
```diff
@@ -13,6 +13,9 @@ def __init__(self,**kwargs):
         :param kwargs:
         """
 
+        self.ratio_residuals = False
+        if 'ratio_residuals' in kwargs:
+            self.ratio_residuals = bool(kwargs.pop('ratio_residuals'))
```
grburgess: Can you make ratio_residuals a private attribute?

threeML/io/plotting/post_process_data_plots.py
```diff
@@ -319,39 +323,45 @@ def display_spectrum_model_counts(analysis, data=(), **kwargs):
...
-                raise NotImplementedError("Not yet implemented")
+                    raise NotImplementedError("Not yet implemented")
 
         residual_plot.add_data(mean_energy,
                                new_rate / new_chan_width,
                                residuals,
+                               residual_yerr=residual_errors,
                                yerr=new_err / new_chan_width,
                                xerr=delta_energy,
                                label=data._name,
```
grburgess: I think we should have a warning that ratios are not statistical quantites when using this option.

volodymyrss: good point! Warning added. Feel free to reword.

Ratios are better than sigmas to asses relative shape of the residual structure, but there is no need to explain it to the user.

giacomov:
@volodymyrss @grburgess can I merge this? It seems we have converged, but I want to make sure before merging.

grburgess:
@giacomov looks good to me. Even if I don't like ratios :)

volodymyrss:
So what's up with this one? It's marked as change requested, and the change was made long ago.

grburgess:
That is my fault. Sorry.

volodymyrss:
Thanks!

PR #201, opened 2017-05-23
test and patch for verifying types of required keywords

volodymyrss:
some spectra might have incorrect types for required keywords.
(apparently some somewhat popular scientific software does this sometimes, I will not say which one)

e.g. if POISSERR is set to 'False' string, instead of boollean, it will trick the whole analysis in thinking the spectrum is poisson.

type checking can be done while reading, suggested implementation in this branch.


threeML/plugins/spectrum/pha_spectrum.py
```diff
@@ -254,14 +257,23 @@ def _read_pha_or_pha2_file(pha_file_or_instance, spectrum_number=None, file_type
...
+            if keyname in _required_keyword_types and type(header.get(keyname)) is not _required_keyword_types[keyname]:
+                warnings.warn("unexpected type of %(keyname)s, expected %(expected_type)s\n found %(found_type)s: %(found_value)s"%
+                        dict(
+                                keyname=keyname,
+                                expected_type=_required_keyword_types[keyname],
+                                found_type=type(header.get(keyname)),
+                                found_value=header.get(keyname),
+                            ))
+            else:
+                gathered_keywords[internal_name] = header.get(keyname)
```
grburgess: Nice! Can you post an example of the output?

giacomov:
Yeah,
unfortunately, OGIP is a standard that nobody really follows, for some obscure reason (most probably because there isn't a validator for it so people don't really check for compliance before publishing their software).

However, if we try to follow all the peculiarities of everybody's implementation of OGIP we will probably end up with an awful and unmaintainable code.

So maybe we can go another route: we can provide a clean implementation which is strictly ogip compliant, but we also provide a new module (`ogip_normalization.py` for example) where we put functions (`normalize_ogip_pha`, `normalize_ogip_rsp`, `normalize_ogip_bak`) which are able to recognize and fix all these small problems (different names, the keyword types, and so on) to make files which are not strictly OGIP-compliant fully compliant. Then our OGIP functions will call as first thing these functions (which will return an open FITS file instance, to avoid writing to disk or modifying the original file) and then proceed assuming a fully-compliant file.

Ideally these `normalize_ogip_*` functions will be made of many calls to other functions like `fix_response_name` or `fix_poisson_keyword` and so on.

What do you think?

threeML/plugins/spectrum/pha_spectrum.py
volodymyrss: There is a test for it, so it's in travis too:

```
  WARNING UserWarning: unexpected type of POISSERR, expected <type 'bool'>
   found <type 'str'>: False
```

grburgess:
@giacomov , something similar to this was suggested by @volodymyrss in a conversation. I think the approach is good as we are already doing this for some special cases (missing TLMIN keywords). We can hopefully put in warnings so that people are aware of the issues in the files. 

From these warnings, we can then suggest writing a corrected file since the corrected one will be living in memory via our PHA class which already has this functionality. What do you think?

volodymyrss:
First of all, I think that the keyword types should be verified, as part of the OGIP standard enforcement. One way or another. What I added in this branch is a rather noninvasive way following the approach which is already implemented. It can be done differently of course.

Then, I do agree with the reasoning that the core OGIP  reader class should implement the reference  OGIP standard. 
I agree the extensions of the standard could be implemented as the normalizing functions.
Though I rather had in mind inheriting from OGIPResponse. 
It could be more flexible: you can read from various strangely organized sources, e.g. databases or generators, as long as they provide MATRIX and EBOUNDS.
OGIPResponse then can write a proper file OGIP. 
Patching files and writing them will produce different results for different source spectra. E.g. INTEGRAL has an extra extension, irrelevant for OGIP.

For example I think there is another very common standard: effectively xspec-readable files. This is a different and loosely defined format (though HDUCLASS business is related to it) and it would be quite useful to have an adapter for it.

It would be great if people followed standards, but the specific case of INTEGRAL has an issue that format is enforced by another internal standard, related to a data access framework, and there is no way it will be changed especially now. Other missions might decide there are their own beautiful frameworks which will interfere..

threeML/plugins/spectrum/pha_spectrum.py
volodymyrss: I was just using it for INTEGRAL, so I pinpoint the examples. I will try to add more with time!

grburgess:
So we just need to decide if we do this quick fix now or do the hard job now.  I do not mind taking this on during the week, but I would need to be pointed to all the possible problems.

volodymyrss:
To be clear, I this branch is just about the fix on types and it follows the logic that is already there.
If there are other ideas on defining and enforcing OGIP keywords and their types, I certainly would not mind to revise..

The first diff does nothing really, just avoiding `except:` (with no exception) which is probably not a good thing. Sorry for mixing it in!

giacomov:
@grburgess @volodymyrss I have merged this for the moment, I guess this can be on the pile of things to do in the future... ;-)

If some of you wants to decide to takes this on (implementing the normalization functions) I'm all for it...

PR #202, opened 2017-05-23
installation instructions case segfault if astromodels is not upgraded

volodymyrss:
coming back to the issue with segfault in astromodels. following the instructions exactly leads to a segfault. 
astromodels should be forcefully upgraded, else an incompatible version pip got as dependency for 3ML remains.

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,7 +206,7 @@ then:
 ```bash
 > pip install numpy scipy ipython
 > pip install git+https://github.com/giacomov/3ML.git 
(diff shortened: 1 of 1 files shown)


giacomov:
good point, thanks!

I am in the process of updating the astromodels release on pip, but this problem will come back in the future as soon as we change astromodels again, so your fix will stay.

I am also preparing a full release on Conda, that will install everything with one command. At that point the pip installation will become less relevant (at least for the general user).

PR #205, opened 2017-05-29
bug importing multinest priors

volodymyrss:
Small change in importing.

(sorry previous pull request was based on wrong master)


grburgess:
thanks for finding this!

giacomov:
Thanks indeed! I am in the process of switching the tests to an anaconda environment which will have multinest available, which means we will introduce tests for multinest and hopefully catch problems like this immediately.

PR #212, opened 2017-06-03
Trigdata

grburgess:
These are some tiny changes to accomadate reading trigdat data in a separate repo of mine.

 I also fixed a bug in the bayesian code where results were not being saved.


grburgess:
@giacomov not sure what is the issue here. I cannot get the coverage to update. We should get this upstream though because MULTINEST is broken without.

PR #222, opened 2017-07-14
found some issues trying to do bayesian analysis

volodymyrss:
trying to make sure it works


giacomov:
Thanks for these fixes!

PR #224, opened 2017-07-16
Spectra plot improve

grburgess:
This moves the SpectrumLike plotting to a member function and allows for over plotting of other models (such as was seen in Vianello et al 2017 ;) ).

The old global function was kept, but is simple a convience wrapper that allows for selecting color sets, etc. 

A few other features were added to improve plotting.


giacomov:
It looks like you are still working on this. Please tell me when you are done and I'll merge it.

Does it change anything regarding the documentation? Please have a look and make sure that the documentation still works.

grburgess:
Ah, I just added some docs. 

Did you add 3 docs for this?

grburgess:
Documentation was added. In the process, I added .from_function class methods to the SpectrumLike classes following the XYLike approach. This should be useful for testing the background modeling in the near future.

grburgess:
@giacomov I am done with this for now. I'll continue to work on the documentation but add that directly .

threeML/plugins/DispersionSpectrumLike.py
```diff
@@ -123,3 +127,121 @@ def write_pha(self, filename, overwrite=False, force_rsp_write=False):
 
         ogiplike = OGIPLike.from_general_dispersion_spectrum(self)
         ogiplike.write_pha(file_name=filename, overwrite=overwrite, force_rsp_write=force_rsp_write)
+
```
giacomov: Nice!

threeML/plugins/SpectrumLike.py
```diff
@@ -234,6 +236,122 @@ def __init__(self, name, observation, background, verbose=True):
         # Apply the mask
         self._apply_mask_to_original_vectors()
 
+    @classmethod
```
giacomov: It looks like this shares a lot of code with the other .from_function method you wrote in this pull request.

Would it be possible to put the common code between the two somewhere else (maybe the parent class)?

grburgess:
@giacomov I removed the duplicated code and added some tests to check fitting of the SpectrumLike classes now that we have generators.

giacomov:
Thanks! Merging it now

Issue #229, opened 2017-07-24
properties of the spectra and choosing sources for SpectrumLike


volodymyrss:
would it make sense to assign some properties in the SpectrumLike plugin. it is quite useful when one wants to associate only some sources to a given spectrum, instead of summing all, as it is done now.

grburgess:
@volodymyrss we have a system for this called 'tags'. This might be the way to address this. For now, tags were to associate plugins with a time interval:

https://github.com/giacomov/3ML/blob/master/threeML/plugin_prototype.py#L11

This is rather old:

https://github.com/giacomov/3ML/blob/master/examples/time_dependent_demo.ipynb

But it may be better to have some other way to do what you are wanting to achieve. Perhaps @giacomov has already a solution coded or in mind.

volodymyrss:
it seems to be called "external_properties"? Ok, it's similar to what I want. I did setters/getters differently though.

giacomov:
@volodymyrss can you describe a bit more the use case for this?

Right now the mindset with which 3ML was designed was to have in the model all sources that are in the sky within a certain region of interest. For non-imaging instruments you had to sum over all the sources, for imaging instruments you treat them one at the time.

However, already for the XYLike plugin, we implemented a solution where you can assign an SED to one specific source in the model. This makes sense for example when you are doing a joint analysis between a large-field instrument and SED data coming from narrow-field instruments.

We can implement something like this for all the OGIP-like plugins as well, if needed.

volodymyrss:
The situation is that I have a number of sources fitted simultaneously, but each source has it's own OGIP. 
Sources share some parameters. If they did not, I would just run the joint fit multiple times, and not have this problem.

But yeah, like XYLike is an options.

Except, more generally, could be not just a source, but also some arbitrary set of properties.
E.g. I have some properties related to detector subsets,  they belong to the data too, and can be used if necessary (e.g. they affect response choice).

Time could be a property too. It seems like it is essentially a property of the data anyway. It can be assigned to IndependentVariable if needed, as it is done in the time-dependent example (by the way, it's broken - several minor inconsistencies following some improvements).

grburgess:
@volodymyrss yes, that example is broken as I am waiting to see what @giacomov does for this case in astromodels. But, we can extend the properties as needed. Also, you can link parameters with astromodels:

http://astromodels.readthedocs.io/en/latest/Model_tutorial.html?highlight=link#linking-parameters

volodymyrss:
[...]

PR #231, opened 2017-07-25
Spectrum source select

grburgess:
Addresses #229 with selecting sources in SpectrumLike and children


giacomov:
@grburgess it looks like there are conflicts here with your own changes in SpectrumLike. Can you take care of that?

giacomov:
@grburgess tests have failed, it looks like there is an indentation error.

grburgess:
@giacomov sorry about that, I was using the editor online and should have just merged manually.

PR #232, opened 2017-07-29
Bkg modeling

grburgess:
Adds background modeling via a plugin.

Cleaned up likelihood calls. 

Made scale factor calculations occur only once as they were being called during every likelihood call.

Added a notebook to the examples that can become a doc when ready.


grburgess:
@giacomov both of these PRs (including #231 ) are passing the tests, but the coverage seems to be broken.

giacomov:
ok, merging now. Thanks!

PR #233, opened 2017-08-03
Gbm daily data

grburgess:
Adds a downloader for GBM daily data

### DIFF
--- a/threeML/plugins/Fermi_GBM/download_GBM_data.py
+++ b/threeML/plugins/Fermi_GBM/download_GBM_data.py
@@ -38,6 +38,77 @@ def _validate_fermi_trigger_name(trigger):
     return trigger
 
 
+_month2number = {'jan':1,
+                  'january':1,
+                  'feb':2,
+                  'february':2,
+                  'mar':3,
+                  'march':3,
+                  'apr':4,
+                  'april':4,
+                  'may':5,
+                  'jun':6,
+                  'june':6,
+                  'jul':7,
+                  'july':7,
+                  'aug':8,
+                  'august':8,
+                  'sept':9,
+                  'september':9,
+                  'oct':10,
+                  'october':10,
+                  'nov':11,
+                  'november':11,
+                  'dec':12,
+                 'december':12}
+
+def _validate_fermi_daily_name(year,month,day):
+
+
+    # check the year
+
+
+
+    assert type(year) == int, 'Invalid year argument'
+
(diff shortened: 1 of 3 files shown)


threeML/plugins/Fermi_GBM/download_GBM_data.py
```diff
@@ -38,6 +38,77 @@ def _validate_fermi_trigger_name(trigger):
     return trigger
 
 
+_month2number = {'jan':1,
```
giacomov: You can use the python datetime module to make sure that the requested date is legal. By doing that you can get rid of almost all the checks. You can also make a better check for a date after Fermi launch. Some examples:

```python

In [13]: fermi_data_start = datetime.datetime(2008,8,4,00,00,00)

In [14]: other_date = datetime.datetime(2008,1,1)

In [15]: other_date > fermi_data_start
Out[15]: False

In [16]: other_date_after = datetime.datetime(2009,1,1)

In [17]: other_date_after > fermi_data_start
Out[17]: True

In [18]: generate_error = datetime.datetime(2009,15,27)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-18-be3df6adb69c> in <module>()
----> 1 generate_error = datetime.datetime(2009,15,27)

ValueError: month must be in 1..12

```

threeML/plugins/Fermi_GBM/download_GBM_data.py
```diff
@@ -273,3 +344,241 @@ def cleanup_downloaded_GBM_data(detector_information_dict):
...
+            if not _warned_tte or data_type=='tte':
+
+                print('Continuous TTE data does not exist for this day')
+
+                _warned_tte = True
+
+    return download_info
+
+
+def _get_latest_version(filenames):
```
giacomov: This looks copy/pasted from download_GBM_data.py, if I am not mistaken. If possible please import it from there (or move it from there to a separate module as a function, and use it in both modules).

We need to avoid copying/pasting identical code as much as possible.

giacomov:
@grburgess what's the status on this?

grburgess:
@giacomov ah, I haven't looked at it in a while. I'll come back to it this week

giacomov:
@grburgess any news on this?

grburgess:
@giacomov I'll have to go back through the code as this is old. I'll update if we still think it is needed.

giacomov:
Closing for inactivity

PR #240, opened 2017-08-18
`HAWCLike.calc_p_value` and bin-dependent top-hat radii for `HAWCLike.display`

sammarinelli:
Added method `threeML.HAWCLike.calc_p_value`, which performs a simple goodness-of-fit calculation for the best-fit model.  Also modified `HAWCLike.display` to allow different top-hat radii for different bins.  Because the `calc_p_value` method relies on functionality recently introduced to AERIE (this morning), this revision may break the build tests unless the test version of AERIE is updated.  Then again it may not break anything since the build test doesn't call the method in question.

### DIFF
--- a/threeML/plugins/HAWCLike.py
+++ b/threeML/plugins/HAWCLike.py
@@ -399,6 +399,19 @@ class HAWCLike(PluginPrototype):
         TS = self._theLikeHAWC.calcTS(self.fitCommonNorm)
 
         return TS
+    
+    def calc_p_value(self, ra, dec, radius):
+    
+        '''
+        Return a p-value for the fit by integrating over a top hat in each bin
+        and comparing observed and expected counts.
+        
+        :param ra: Right ascension in degrees of top-hat center.
+        :param dec: Declination in degrees of top-hat center.
+        :param radius: List of top-hat radii in degrees (one per analysis bin).
+        '''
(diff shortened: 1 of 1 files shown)


giacomov:
@sammarinelli thanks for this contribution. Could you please elaborate a bit on the method used to compute the p-value?

sammarinelli:
@giacomov The p-value is calculated by computing D = 2 ln[(saturated likelihood)/(best-fit likelihood)], where the "saturated model" is the model corresponding to setting each bin's expectation equal to its observed count.  The quantity D should then obey a chi^2 distribution with (number of bins) - (number of free parameters) degrees of freedom under the hypothesis that the fit is good, so the p-value is calculated by integrating the chi^2 distribution from D to infinity.  The bin contents are computed by integrating over a top hat rather than pixel by pixel to ensure there are good enough statistics to apply Wilk's theorem.  I didn't try to set this up to interact with `threeML/classicMLE/goodness_of_fit.py` because I believe that module tries to do the calculation using Monte Carlo rather than analytically.

giacomov:
I implemented the changes requested here and also implemented the relative tests in the master branch. I am closing this. 

thanks for the contribution!

sammarinelli:
Thanks!

PR #248, opened 2017-09-24
Add significance per interval

grburgess:
Just a small convenience property for getting the significances of intervals after they are binned.


grburgess:
@giacomov do you know why code coverage is having so many problems? There is no evenflow.

PR #254, opened 2017-10-06
Move ogip

grburgess:
moved files out of the plugins folder as they are utils.


grburgess:
@giacomov this should be working now. Basically, I moved things outside of the plugins directory to the utils directory. Feels cleaner!

grburgess:
I'm merging. We may need to update some examples

PR #265, opened 2018-01-14
Astroq update

grburgess:
move to new astroquery library for catalogs


grburgess:
This is a simple merge... going for it!

PR #266, opened 2018-01-14
Add error out

grburgess:
allows for getting sigmas from RVs.


threeML/random_variates.py
```diff
@@ -117,6 +118,77 @@ def highest_posterior_density_interval(self, cl=0.68):
 
         return hpd_left_bound, hpd_right_bound
 
+    def highest_posterior_density_sigma(self, cl=0.68, allow_asymmetric=False):
```
giacomov: I would remove this method and the next.

* There is almost 100% code duplication
* It will fail most of the time with the default `allow_asymmetric=False` because low_bound and hi_bound are going to be slightly different most of the time even when the likelihood is gaussian because of limited sampling precision. 

I would instead add the `interval_to_errors` method to the general `__init__.py` and ask the user to explicitly do:

```
interval_to_errors([....].highest_posterior_density())
```

(or any other interval yielding method).

grburgess: Agreed.

PR #268, opened 2018-01-14
Plotting fixes

grburgess:
Allows for linear scales in spectrum model+data plots


threeML/io/plotting/post_process_data_plots.py
```diff
@@ -118,6 +121,41 @@ def display_spectrum_model_counts(analysis, data=(), **kwargs):
     if 'show_residuals' in kwargs:
         show_residuals= bool(kwargs.pop('show_residuals'))
 
+    if 'show_model' in kwargs:
```
giacomov: I'd suggest using the normal default argument way of python instead of `**kwargs`:

```python
display_spectrum_model_counts(analysis, data=(), 
                              step = True, 
                              data_cmap = threeML_config['ogip']['data plot cmap'],  # plt.cm.rainbow
                              model_cmap = threeML_config['ogip']['model plot cmap'],  # plt.cm.nipy_spectral_r
                              show_legend = True,
                              show_residuals = True,
                              with_model = True, 
                              data_colors = cmap_intervals(len(data_keys), data_cmap),
                              model_colors = cmap_intervals(len(data_keys), model_cmap),
                              .... ):
    ....
```
so that you can avoid most of those `if/else` statements and make the code a lot shorter (and a bit cleaner).

For those arguments that you need to build programmatically from other things (like `model_labels`) use a default value of `None` in the calling sequence and test it as you already do:

```python
if model_labels:
   # Make sure the provided labels have the desired length
    assert len(model_labels) == len(data_keys), 'you must have the same number of model labels as data sets'
else:
   # Get one label for each dataset
    model_labels = ['%s Model' % analysis.data_list[key]._name for key in data_keys]
```

grburgess: I agree, we were just getting a lot of issues in code climate for too many function arguments. I'll fix this up.

grburgess: @giacomov I think it is really dangerous to have variables as default args, no?

giacomov: What do you mean by ’variables’? And why is it dangerous?

grburgess: https://google.github.io/styleguide/pyguide.html#Default_Argument_Values

I'm not sure we should put in the config as default args.

giacomov: I was thinking earlier, maybe the best solution would be to define `styles` a bit like seaborn does (or CSS for HTML). 

What I envision is an use like this (from the user perspective):

```python
with plot_style('plain'):
    
   jl.display_spectrum_model_counts(...)
   jl.some_other_plotting_command(...)
```
where all keywords pertaining to the style of the plot are not part of the calling sequence.

This will mean avoiding several repetitions of the same keywords in multiple functions, as well as giving a way for the user to produce easily a set of plots which are coherent in terms of colors and styles and to experiment quickly by just changing the style.

The style needs to have these features:

* There is a default style (`plain`) defined in a yaml file `plain.yml` which is active by default, and saved in a `styles` subdirectory of the 3ML config dir (where the main config file is)
* The user can customize the style any way he/she wants during a session, and save it with a custom name (say `mystyle`, and under the hood it will be saved in a file `mystyle.yml`)
* in a following session he/she can use the style using a context manager like shown above (the `with` statement)
* in the main config file he/she can specify the name of the default style. It will be `plain` by default, but it can be changed to anything

All in all, it means just saving/loading in YAML files instances of `collections.OrderedDict` corresponding to the various styles, and implementing a simple context manager.

How does it sound?

I can implement this easily (or you can if you like).

PR #270, opened 2018-01-15
Simplify spec likelihood

grburgess:
I've tried to simplify the SpectrumLike handling of different likelihoods by making an object that has the same call regardless of the noise model. This reduces the complexity of the code in the likelihood call and simulation and gives a little speed bump.

It also makes maintenance of the code easy in the future as we do not have to keep track of switches and instead just add new likelihoods on as needed. 

There are lots of little changes as I moved things around.


giacomov:
@grburgess can you please solve the conflict here?

grburgess:
@giacomov got it!

threeML/plugins/SpectrumLike.py
```diff
@@ -15,26 +15,27 @@
...
 NO_REBIN = 1E-99
 
 __instrument_name = "General binned spectral data"
 
 # This defines the known noise models for source and/or background spectra
-_known_noise_models = ['poisson', 'gaussian', 'ideal']
+_known_noise_models = ['poisson', 'gaussian', 'ideal', 'modeled']
 
 
 class SpectrumLike(PluginPrototype):
```
giacomov: The constructor is quite long with a lot of `if/else` and Codecov does not like it. I suggest factoring it in multiple shorter methods. For example, lines 93-230 can be placed in a method like  `_probe_noise_models`. If you decide to go down this way, please do not assign members of the class from within the new method, but return them instead. For example:

```python
# WRONG
def __init__(self):
    self._method()

def _method(self):
    self._member1 = 'foo'
    self._member2 = 'bar'

# CORRECT
def __init__(self):
    self._member1, self._member2 = self._method()

def _method(self):
    member1 = 'foo'
    member2 = 'bar'
    return member1, member2

File: threeML/plugins/SpectrumLike.py [hunk 2]
```diff
@@ -264,6 +268,8 @@ def __init__(self, name, observation, background=None, verbose=True, background_
 
         if self._background_plugin is not None:
 
+            self.background_noise_model = 'modeled'
+
             for par_name, parameter in self._background_plugin.likelihood_model.parameters.iteritems():
```
giacomov: `iteritems()` is not supported in python 3. I know that we use it in many places, and we'll need to fix that, but for all new code it is better to use:

 ```python
for key, value in dictionary.items():
    # Do whatever
```

when the dictionary is short (< thousands), and the following when the dictionary is going to be long:

```python
for key in dictionary:
    item = dictionary[k]
```

Both constructs work in py2 and py3, although the first one can be memory intensive in py2 on large dictionaries (which is why the 2nd construct is to be preferred in those cases).

threeML/utils/spectrum/spectrum_likelihood.py
```diff
@@ -0,0 +1,315 @@
...
+import numpy as np
+
+
+
+
+# These classes provide likelihood evaluation to SpectrumLike and children
+
+
+
+class SpectrumLikelihood(object):
```
giacomov: very nice!

threeML/utils/spectrum/spectrum_likelihood.py
```diff
@@ -0,0 +1,315 @@
...
+            randomized_background_counts[idx] = 0
+
+        return randomized_background_counts
+
+    def get_randomized_background_errors(self):
+
+        return copy.copy(self._spectrum_plugin.background_count_errors)
+
+
+likelihood_lookup = {'poisson':{'poisson' : PoissonObservedPoissonBackgroundLikelihood,
```
giacomov: If the user wants to add its own statistic, he/she can subclass the SpectrumLikelihood class. But then, how can he use the new statistic? I guess he/she will need to add it to this dictionary, but then how is he going to set it for the spectrum?

threeML/utils/spectrum/spectrum_likelihood.py
giacomov: although, maybe `BinnedStatistic` is a more appropriate name? `Likelihood` could be confused with `Like` that we use for plugins.

threeML/utils/spectrum/spectrum_likelihood.py
```diff
@@ -0,0 +1,315 @@
...
+        
+        A class to hold the likelihood call and randomization of spectrum counts
+        
+        :param spectrum_plugin: the spectrum plugin to call
+        """
+
+        self._spectrum_plugin = spectrum_plugin
+
+
+    def get_log_likelihood(self):
```
giacomov: `get_statistic` or even `get_current_value` might be more appropriate, otherwise, this method has a name very similar to what is used in the plugins to get the likelihood value.

threeML/plugins/SpectrumLike.py
grburgess: I agree, we've been reducing it down overtime. It seems we keep coming back to this one :) I'll do some work on it and commit here.

threeML/plugins/SpectrumLike.py
grburgess: Ok, I've been thinking about these little issues as well. We might also what to take a look at how we deal with strings as this is different in py3

threeML/utils/spectrum/spectrum_likelihood.py
grburgess: yeah, I thought about that one. Will update.

threeML/utils/spectrum/spectrum_likelihood.py
grburgess: 👍

threeML/utils/spectrum/spectrum_likelihood.py
grburgess: I can think of a way to do this. Sort of like how you can add models in astromodels. However, I don't personally like the idea of custom likelihoods. This is the problem with XSPEC. I understand that we allow for the changing of the noise models (we should add a warning!), but allowing too much freedom and we get 3ML Chandra papers with bonkers statistics.

giacomov: I understand your point, but we should also allow the advanced users the freedom they deserve. Say that somebody needs a statistic based on the Skellam distribution... or maybe somebody wants to experiment with some custom analysis and they need some weird statistic...

When we added PGstat in Xspec, I had to personally write Keith Arnaud and ask him to implement it. He was very kind and did it in a very reasonable time frame, but I don't think this is optimal.

grburgess: I agree. It's open source, so there is always the chance. I think a good compromise would be to to retain the original data's statistical properties and warn the user pretty blatantly that the choice being made is not reflecting the data's distribution.

threeML/plugins/SpectrumLike.py
grburgess: Um. Is the setting of class members outside of the constructor an issue with pickling? Can you point me to some guidelines? This becomes a little cumbersome in tis case because there are a lot of things set in nested logic.

grburgess:
Ok. I fixed all the above issues. As we discussed in private, I'm leaving the BinnedStatistic classes static at the moment. There are two options add a new statistic. The user can subclass The plugin and modify the get_log_like. This is actually not so difficult. The other option is to modify the code to have the appropriate noise models known. Either way, adding a statistic is something that should be thought out and I would expect any user that is that advanced to be able to deal with the concept in the first place. We can also be contacted

PR #272, opened 2018-02-04
fixed non-updating intervals

grburgess:
Fits to time series backgrounds were not properly updating the intervals when the user selected time intervals that were outside of the data ranges. This caused time bins with ZERO counts to be included in the fit. This would have resulted in a stall during the fit or incorrect fits. This has now been fixed.


grburgess:
Fixes #273

PR #280, opened 2018-03-01
Multinest test

grburgess:
I did not commit as you may be switching the test directory. 

I am also working on the BayesianAnalysis class as there are some issues with a few of the methods.


grburgess:
@giacomov I have also added to the codecov the ability to ignore the test directory and the tutorial material. This may help us to avoid having to move these files.

giacomov:
@grburgess can you rebase (i.e., get the latest changes from master) and see if the Travis build succeed? I already solved the problems with the two tests failing, you probably forked before that was ready.

PR #299, opened 2018-08-02
Add cstat bkg to timeseries

grburgess:
This allows for the extraction of the selected background interval's counts rather than using the polynomial modeled counts. The choice comes at the point of creation of the plugin. All the identifiers  in the background spectrum are set automatically and the background exposure is properly scaled.


grburgess:
Something strange is happening with the OSX builds

giacomov:
I already fixed it, please rebase and let's see if that solves the build here

grburgess:
Shall we merge this as well?

PR #302, opened 2018-08-27
Set active measurements list

ColasRi:
Add a `bin_list` option to set_active_measurements(), similar to the behavior in the hawc_hal plugin.


ColasRi:
There seem to be a bug with codecove, I increase the coverage by 0.33% (all new lines are covered), not decrease it.

PR #310, opened 2019-01-04
Sorting tte data

grburgess:
I found some bugs in TTE data which caused Bayesian blocks to fail At first I thought it was just sorting, but it is actually that a lot of files have duplicate time tags which means the files are corrupt.

I've now added a check for duplicate time tags which will cause an assertion error. If there are no duplicates and the data are still not sorted, a warning is issued and we sort the time for the user.

### DIFF
--- a/threeML/utils/binner.py
+++ b/threeML/utils/binner.py
@@ -528,7 +528,7 @@ class TemporalBinner(TimeIntervalSet):
 
             if 'duplicate' in str(e):
 
-                warnings.warn('There where possible duplicate time tags in the data. We will try to run a different algorithm')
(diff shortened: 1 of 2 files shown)


giacomov:
yes,the new version of `pytest` breaks our tests.

I am in the process of revamping the whole testing system, it should fix these and other problems. In the meantime, you can commit a change to the `meta.yaml` file that selects the previous version of pytest (`3.10.*`)

grburgess:
@giacomov  do you mean the travis yaml file?

giacomov:
I just committed the change to master. You can rebase and you should get the fix.

PR #314, opened 2019-02-05
updated comment on giacomos significance formula

grburgess:
Adding a doc string to the GV significance calculation


omodei:
Any idea how to fix the Travis CI? It looks like it fails to find the "toolchain" module.

grburgess:
@omodei errr, I do not know at the moment. It's probably some random version thing that pops up every now and then. I'll have a look in the moring before merging. Thanks for looking at the PR!

grburgess:
@omodei I still can't figure this out. Should we commit this merge (comments only) or wait until we solve the issue?

Issue #317, opened 2019-03-07
FermiLATSourceCatalog fails 


omodei:
Due to the new release of the 4FGL, the schema in the FermiLATSourceCatalog (catalogs/Fermi.py) is no longer valid.
lat_catalog = FermiLATSourceCatalog()
ra, dec, table = lat_catalog.search_around_source("Crab", radius=20.0)
crashes

grburgess:
Thanks for catching this! Too bad our tests keep failing for other reasons as this might have been flagged in the cron builds. I'll review when you have a fix.

omodei:
This issue is fixed.

PR #318, opened 2019-03-07
Flpsc

omodei:
The new 4FGL catalog has some differences in format from the 3FGL, in particulars the names of the parameters of the models, and the "tvecat_assoc" column. This caused the code in Fermi.py to crash. The changes I have implemented are NOT backward compatible, but the VO interface to the fermi catalog automatically retrieve the 4FGL, so there is no way to use the 3FGL catalog, which is, by all means, superseded by the 4FGL.


grburgess:
Addresses #317

grburgess:
Hi @omodei this looks like it is working. I went ahead and did a PR as you might be on a plane. I added in the  master and the tests are working. I did a quick review and it all looks cool. Can merge for your tutorial when ready!

grburgess:
Addresses #317

Issue #305, opened 2018-10-05
Plot XYLike data points on fit result


colasri:
It would be nice to have the possibility to plot the XY data points of the XYLike class on the same plot as the `threeML.plot_point_source_spectra()`.
The `XYLike.plot()` method does not provide this since it's a very simple method not dealing with units, nor multiply by an E^2 factor for the Y scale. Any idea on how to best achieve this @grburgess ?

grburgess:
Hi sorry for the delay! I can have a look at how to solve this. Ideally, this is *not* a good idea as the x-axis between the plugins is not guaranteed to be the same. SpectrumLike could have _reconstructed_ energy that is highly instrument dependent, and XYLike could have _true_ energy. Thus, it would convey an imprecise understanding of these two data spaces. That being said, I know how to fix it and will address it now :)

PR #324, opened 2019-03-29
add the ability to pass an axes instance to the plotting function

grburgess:
Addresses the request in #305

### DIFF
--- a/threeML/plugins/XYLike.py
+++ b/threeML/plugins/XYLike.py
@@ -399,25 +399,46 @@ class XYLike(PluginPrototype):
 
         return new_xy
 
-    def plot(self, x_label='x', y_label='y', x_scale='linear', y_scale='linear'):
+    def plot(self, x_label='x', y_label='y', x_scale='linear', y_scale='linear', ax=None):
+        """
+        Plot the data and model if a model has been set. 
 
-        fig, sub = plt.subplots(1,1)
+        :param x_label:  the label of the x-axis 
+        :param y_label: the label of the y-axis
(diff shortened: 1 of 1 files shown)


threeML/plugins/XYLike.py
```diff
@@ -399,25 +399,46 @@ def _new_plugin(self, name, x, y, yerr):
...
 
         if self._likelihood_model is not None:
 
             flux = self._get_total_expectation()
 
-            sub.plot(self.x, flux, '--', label='model')
+            ax.plot(self.x, flux, '--', label='model')
 
-            sub.legend(loc=0)
+            ax.legend(loc=0)
```
henrikef: Should we make the legend optional in case we are supplying an axes object? In that case, the user might want to handle the legend themselves.

threeML/plugins/XYLike.py
```diff
@@ -399,25 +399,46 @@ def _new_plugin(self, name, x, y, yerr):
...
+
+            fig = ax.get_figure()
+            
 
-        sub.errorbar(self.x, self.y, yerr=self.yerr, fmt='.')
+        ax.errorbar(self.x, self.y, yerr=self.yerr, fmt='.')
 
-        sub.set_xscale(x_scale)
-        sub.set_yscale(y_scale)
+        ax.set_xscale(x_scale)
```
henrikef: Should we make setting the scale and axis labels optional for the case that the user passes their own axes object? I'm thinking to just have the default as None for x_label, x_scale etc and only change them if something other than None is provided by the user?

PR #346, opened 2020-04-07
Jmb py2to3

ndilalla:


grburgess:
My local tests worked out and I did some of the commits so I think I cannot be trusted there. So I am giving the **all go** before I go to sleep. Amazing work @ndilalla ! It Is great to have you on the tea,

ndilalla:
> My local tests worked out and I did some of the commits so I think I cannot be trusted there. So I am giving the **all go** before I go to sleep. Amazing work @ndilalla ! It Is great to have you on the tea,

Thanks @grburgess ! :smiley: You did a great job here. Let's merge this one and see what happens.

grburgess:
Is it ready?

ndilalla:
I think yes. Unfortunately it is not compliant with root6 yet, but we are going to work on it in a separate branch. The only thing I didn't test is the installation via cron job but I can't do it without uploading before the new version on conda. So, in case, a quick fix might be needed.

grburgess:
Also, ROOT7 is on the way... let's merge.

Issue #352, opened 2020-04-14
OGIPLike trying to read missing background file when keyword is set to ' '


domeckert:
Hi

Trying to read a PHA file produced by XSPEC with no background ("fakeit none") I got the following error:

ValueError: Empty filename: ''

It appears it is trying to read an empty background file. When setting the BACKFILE keyword to NONE the problem disappears.

grburgess:
Thanks for finding this. It should now be solved!

PR #354, opened 2020-04-14
fix loading of background when file is empty string

grburgess:
if backfile existed but was empty, it crashed OGIP. This fixes that.


grburgess:
fixes #352

PR #368, opened 2020-08-05
Add functionality to FermiLatLike plugin

xboluna:
Allows FermiLatLike plugins to be given a single source from a model. Allows for multiple sources in different ROI to be fit conjunctively without expecting each other in the same background.

Source name can be set at initialization, or when calling set_model.


ndilalla:
Hi @xboluna, great job! May I ask you (if it's not already there) to add a simple test for these new features? Something very basic. Thanks!

grburgess:
Thanks! this looks really cool. All your contributions are much appreciated! I'm happy to merge, but as this is LAT, the experts should make the final call @omodei and @ndilalla

threeML/plugins/FermiLATLike.py
```diff
@@ -277,31 +300,45 @@ def __init__(
         # Activate inner minimization by default
         self.setInnerMinimization(True)
 
+        if source_name is not None:
```
omodei, line 303: Is this 'if statement really needed? It looks like you can always set self._source_name = source_name

xboluna: Good point. Thanks!

Issue #372, opened 2020-08-24
Cannot display OGIP-like (XMM pn) spectra


ivvv:
I am trying to use threeML with XMM data. Reading the OGIP-like spectra (PHA, background, ARF, RMF) seems to work OK but when I try to display the counts spectrum it fails. Here is some code:

```
import os

os.environ['OMP_NUM_THREADS'] = "1"
os.environ['MKL_NUM_THREADS'] = "1"
os.environ['NUMEXPR_NUM_THREADS'] = "1"

from threeML import *

pha = "P0134531601PNU002SRSPEC0001.FTZ"
#
# no need to provide the BACK, ARF and RMF files as they are given in the header of the PHA
#  
#bak = f"{wdir}/P0134531601PNU002BGSPEC0001.FTZ"
#arf = f"{wdir}/P0134531601PNU002SRCARF0001.FTZ"
#rsp = f"{wdir}/epn_e1_ff20_sdY8.rmf"

#ogip = OGIPLike("ogip", pha, background=bak, response=rsp, arf_file=arf)
ogip = OGIPLike("ogip", pha)

fig = ogip.view_count_spectrum()
```

this fails with the following error:

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
 in 
      4 #ogip.set_active_measurements('0.3-12.0')
      5 ogip.rebin_on_background(5)
----> 6 fig = ogip.view_count_spectrum()

~/miniconda3/envs/threeML/lib/python3.7/site-packages/threeML/plugins/SpectrumLike.py in view_count_spectrum(self, plot_errors, show_bad_channels, show_warn_channels, significance_level, scale_background)
   2226                 # convert to rates, ugly, yes
   2227 
-> 2228             background_counts /= self._background_exposure
   2229             background_errors /= self._background_exposure
   2230 

TypeError: ufunc 'true_divide' output (typecode 'd') could not be coerced to provided output parameter (typecode 'l') according to the casting rule ''same_kind''
```
I have tried with rebinning the background, rebinning the source or limiting the energy ranges and it still does not work.

```
ogip.set_active_measurements('0.3-12.0')
ogip.rebin_on_background(5)
```

The 4 test files (as produced by the XMM standard pipeline) are available in here:
ftp://ftp.sciops.esa.int/pub/ivaltchanov/export/xmm_pn_3ml_data_test.zip

grburgess:
I've pushed a fix that should fix the issue.

[...]

PR #373, opened 2020-08-24
fix int conversion

grburgess:
Addresses the bug in #372 

Issue was trying to convert a numba array that was set as an int to a float...


omodei:
This seems reasonable. Not fully understood why you need old_div, and not simply /, but I guess it is for backward co potability with python 2 (although there should not be any division by integers...)

grburgess:
the old div preserves non integer division.

PR #383, opened 2020-10-29
Fix test catalog

ndilalla:
This PR fixes the problems with ʻastropy` 4.1+ (which caused catalog tests to fail) and tries to help conda solve the py27 environment on travis.


grburgess (approved):
This looks like a smart way to deal with this!

PR #386, opened 2020-11-05
Ar hdf5

grburgess:
Adds the ability to save AnalysisResults to HDF5 files which are more flexible than FITS, thus we can expand on their content in the future. Conversion utilities are supplied to convert old saved files in FITS format. 

Does not break past result saving.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,12 +8,6 @@ env:
 jobs:
   include:
   - os: linux
-    python: 2.7
-    env: TRAVIS_PYTHON_VERSION=2.7
-  - os: osx
-    python: 2.7
-    env: TRAVIS_PYTHON_VERSION=2.7
-  - os: linux
     python: 3.7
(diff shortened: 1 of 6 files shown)


grburgess:
looks like some this failed with python2...

henrikef:
Do we still need to support python 2 at this point?

ndilalla:
I will look into this and see if we can continue to support py27.

grburgess:
I am happy to drop py2 support at this point.

ndilalla:
@grburgess I dropped py27 support. Do you have any idea why it is now failing on py37 (linux)?

grburgess:
@ndilalla I am checking.

grburgess:
@ndilalla Ok, I fixed it... another one of these very strange byte string issues.

ndilalla:
> @ndilalla Ok, I fixed it... another one of these very strange byte string issues.

Great, thank you! Now I can merge this one and tag the first py37-only release.

Issue #385, opened 2020-11-03
Incompatibility(?) with fermipy 1.0.0?


henrikef:
I'm seeing some issues with `test_FermipyLike`, which is skipped in the travis ci tests.

I'm using theeML 2.0.1, astromodels 2.0.1, fermitools 2.0.0, fermipy 1.0.0 with python 3.7.8, all installed via conda.

All tests pass except for `test_FermipyLike`, which at first gave me an `AssertionError` as it found more sources in the LAT catalog than expected (see [log1.txt](https://github.com/threeML/threeML/files/5483457/log1.txt)). That should be an easy fix, I assume this is due to the change from 4FGL to 4FGL-DR2?

However, after adjusting the number of sources in the test, I'm getting a different error related to `tmin` and `tmax` being set to `None` in the fermipy config (see [log2.txt](https://github.com/threeML/threeML/files/5483458/log2.txt)). There are also a lot of user warnings about setting a lower bound on the source normalization which are annoying, but not too concerning.

Any ideas? I'm not an expert on fermipy unfortunately.

PR #390, opened 2020-12-08
Fix to support fermipy 1.0.0.

ndilalla:
This pull request implements the fix to support the new version of `fermipy` (issue #385). Unfortunately the plugin is still not working because of a couple of issues on the fermipy side. For reference these are the related issues that have to be solved:
- https://github.com/fermiPy/fermipy/issues/369
- https://github.com/fermiPy/fermipy/issues/372

### DIFF
--- a/threeML/plugins/FermipyLike.py
+++ b/threeML/plugins/FermipyLike.py
@@ -4,6 +4,7 @@ import astromodels
 import numpy as np
 import os
 import yaml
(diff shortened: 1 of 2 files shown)


omodei (approved):

henrikef:
Looks good! It's unfortunate we can't test the full analysis due to the bugs in fermipy.

henrikef (approved):

PR #395, opened 2020-12-18
New features

BjoernBiltzinger:
Add/Improve a few features:

Major:

1. Add an option "share_spectrum" to the bayesian base_sampler. If this is set to True, it is checked in the beginning, if several detectors in the data_list have the same input Energy-bins. If yes, we can use this information to only calculate the spectrum once and distribute to the different detectors that use the same input Energy-bins. Can give a huge speed improvement if the spectrum calculation is slow and many similar detectors are used (like for GBM or SPI). Only works for SpectrumLike and DispersionSpectrumLike.
Can be used like this: 
ba = BayesianAnalysis(model, datalist)
ba.set_sampler("multinest", share_spectrum=True)

2. The integral function in SpectrumLike calculates the spectrum for every e1, e2 and (e1+e2)/2 (e1 lower ends and e2 upper ends of all input Energy-Bins) for the simpson integration. But, as this is the integration on the input side, there will be no gaps in the energy bins, therefore e2[i-1]=e1[i] => The spectrum is calculated two times for the same energy value. This was changed to just calcualte the spectrum at every Ebin-Edge (and also still the middle Energy-Bin values) and therefore gives a speedup of ~ 1/3 for the integration of the incoming flux.

3. Give the posibility to also use a trapz integration instead of the simpson integration. Less accurate, but needs only have the number of spectrum evaluations. -> speed up of ~ 1/2 for the integration of the incoming flux. 
Can be changed with SpectrumLike.set_model_integrate_method("trapz") for the source model integration or SpectrumLike.set_background_integrate_method("trapz") for the background spectrum integration. Simpson integration is still the 
default.

Plotting:
1. Fix cmap input for the display_spectrum_model_counts function.
2. Add an option "data_per_plot" to display_spectrum_model_counts, with which one can specify how many detetors should be plotted in one plot. If num_detectors>data_per_plot, the plotting is splitted into several plots.
3. Add an option "show_background" to display_spectrum_model_counts/display_model that will additionally plot the background rate.
4. Add an option "source_only" to display_spectrum_model_counts/display_model. If this is set to True, the "data" that is plotted, is observed_data-background_model and the model is only the source model (This is how it is done up to now). If this is set to False, the data is the observed_data and the model is the source_model+background_model.

Minor:
1. Added a few typings for the things I changed.
2. Changed the construct_counts_array method in SpectrumLike to store a few more information, like for example the background rate (needed for the new plotting)
3. Change the restore_median_fit function for bayesian sampler. Do not set every parameter to the median of the individual posterior distributions, but rather set the parameters to the sample with max loglike.
5. Add test for the share_spectrum, improve tests for display_spectrum_model_counts, to cover more of the optional input.
4. Small change in BinnedSpectrumSeries to make it possible to construct it from BinnedSpectra with no dispersion.

### DIFF
--- a/threeML/bayesian/sampler_base.py
+++ b/threeML/bayesian/sampler_base.py
@@ -29,10 +29,13 @@ from threeML.analysis_results import BayesianResults
 from threeML.utils.statistics.stats_tools import aic, bic, dic
 from threeML.exceptions.custom_exceptions import LikelihoodIsInfinite, custom_warnings
 from astromodels.functions.function import ModelAssertionViolation
-
+from threeML.plugins.SpectrumLike import SpectrumLike
+from threeML.plugins.DispersionSpectrumLike import DispersionSpectrumLike
+from threeML.data_list import DataList
+from astromodels.core.model import Model
 
 class SamplerBase(with_metaclass(abc.ABCMeta, object)):
-    def __init__(self, likelihood_model, data_list, **kwargs):
+    def __init__(self, likelihood_model : Model, data_list : DataList, **kwargs):
         """
 
         The base class for all bayesian samplers. Provides a common interface
@@ -40,7 +43,9 @@ class SamplerBase(with_metaclass(abc.ABCMeta, object)):
 
 
         :param likelihood_model: the likelihood model
-        :param data_list: the data list 
+        :param data_list: the data list
+        :param share_spectrum: (optional) Should the spectrum be shared between detectors
+        with the same input energy bins?
         :returns: 
         :rtype: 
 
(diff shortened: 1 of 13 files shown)


threeML/bayesian/sampler_base.py
```diff
@@ -57,7 +62,62 @@ def __init__(self, likelihood_model, data_list, **kwargs):
...
+            num_found = 0
+            self._data_ein_edges = {}
+            for j, d in enumerate(list(self._data_list.values())):
+                if j == 0:
+                    if isinstance(d, DispersionSpectrumLike):
+                        self._data_ein_edges[num_found] = d.response.monte_carlo_energies
+                    elif isinstance(d, SpectrumLike):
+                        self._data_ein_edges[num_found] = d.observed_spectrum.edges
+                    else:
+                        raise AssertionError("Share spectrum only works for"\
```
grburgess, line 87: Why not just skip those that are not correct?

threeML/plugins/SpectrumLike.py
```diff
@@ -1737,22 +1763,64 @@ def differential_flux(energies):
...
+                        (e2 - e1)
+                        / 6.0
+                        * (
+                            differential_flux(e1)
+                            + 4*differential_flux((e2 + e1) / 2.0)
+                            + differential_flux(e2)
+                        )
+                    )
+
+        elif integrate_method == "trapz":
```
grburgess, line 1804: we should add this as a configuration option

threeML/test/conftest.py
```diff
@@ -75,6 +75,21 @@ def get_dataset():
 
     return NaI6
```
grburgess, line 77: please do not use CamelCase and switch to pathlib here :)

grburgess (changes requested):
Just a few comments

grburgess (approved):

PR #403, opened 2021-01-02
Faster photo

grburgess:


grburgess:
I couldn't comment because there were so many files changed that it crashed the webpage. 

These are all the old CSV files that we stored for filters. They are now gone and replaced with a single hdf5 file. This also eliminates the annoying print out when installing and will result in smaller diffs. 


The interface to PhotometryLike also changed. There is now a container class for photometric observations which can be serial to HDF5... we could likely interface to other optical formats... but those are extremely varied. 

The likelihood ism now 15-20x faster by getting rid of the speclite convolution which relied on units and doing our own which should be future proof.

Issue #392, opened 2020-12-16
Update to Iminuit 2.0 interface. 


grburgess:
iMinuit made breaking changes to their interface:

https://iminuit.readthedocs.io/en/stable/changelog.html

This needs to be updated, and if further improvements can be made... it would be great. I will leave it open to who would like to do this, but @cescalara has shown interest. 

@giacomov I am only putting you there if you still love Minuit and because I think they have now updated the minimizer...

PR #409, opened 2021-01-23
Update iminuit interface to >=v2.0

cescalara:


grburgess:
Address #392

threeML/minimizer/minuit_minimizer.py
```diff
@@ -272,7 +226,7 @@ def _compute_covariance_matrix(self, best_fit_values):
 
         try:
 
-            covariance = np.array(self.minuit.matrix(correlation=False))
+            covariance = np.array(self.minuit.covariance)
```
grburgess, line 229: Is there a reason to cast to array here?

cescalara: Otherwise the type is `iminuit.util.Matrix`, so was trying to be consistent.

grburgess (approved):

PR #411, opened 2021-01-28
Feature omegaconf

grburgess:
This rewrites how our configuration is handled by  moving from our custom solution to one derived from OmegaConf. 
Nothing changes as far has how the old code worked except that a few of the things have moved or replaced spaces with underscore. Also, the *default* location of the config has moved from .threeml to .config/threeml to be more consistent with proper *nix practice. However, you can change this with export THREEML_CONFIG="some_directory" while will be scanned for custom yaml files.

What this brings:

-  Ability to have partial custom configs. In the config directory, you can have one of more yaml files that change only the things you want to without having to hold a copy of the whole configuration file. 
- The location of the configs can be changed
- config variables can be set with ENV variables by substituting ${env:<VAR>} where <VAR> is whatever you want... e.g. a path
- dot and key access to the config for tab completion 
- the default config is built from classes making some things reusable and easier to manage
- a function to copy the CURRENT configuration to disk if you modified during a session
- a display function which also lets you pass sub menus (catalogs for example)
image

I think this will be easier to maintain (this is a pro library) and easier to build upon in the future rather than having to keep track of deep spacing in a yaml file.


conda-dist/recipes/threeml/meta.yaml
```diff
@@ -55,6 +55,8 @@ requirements:
     - interpolation>=2.1.5 # [py3k]
     - tqdm
```
BjoernBiltzinger, line 56: tqdm=4.56.0 was needed for me to work with the colored progress bars

BjoernBiltzinger (commented):
Very nice. I only see two minor things.

threeML/config/config_structure.py
```diff
@@ -0,0 +1,65 @@
+import logging
+from dataclasses import dataclass, field
+from enum import Enum, Flag, IntEnum
+from typing import Any, Dict, List, Optional
+
+from omegaconf import II, MISSING, SI, OmegaConf
```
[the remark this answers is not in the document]
grburgess: Ah, correct... this was just there when I was building things.

conda-dist/recipes/threeml/meta.yaml
grburgess: Thanks!

henrikef (approved):
Looks good to me (although we should merge my branch with the fixes for iminuit into this one first).

PR #421, opened 2021-02-04
Anastasia Tsvetkova

anastasia-tsvetkova:
Class method 'from_konus_pha' necessary to read KW PHA2 has been added.


anastasia-tsvetkova:
Class method 'from_konus_pha' necessary to read KW PHA2 has been added.

PR #425, opened 2021-02-09
Feature fast flux

grburgess:
Some doc fix ups and a faster way to compute fluxes.

### DIFF
--- a/docs/md_docs/flux_examples.md
+++ b/docs/md_docs/flux_examples.md
@@ -1,0 +1,273 @@
+---
+jupyter:
+  jupytext:
+    formats: ipynb,md
+    text_representation:
+      extension: .md
+      format_name: markdown
+      format_version: '1.2'
+      jupytext_version: 1.7.1
+  kernelspec:
+    display_name: Python 3
+    language: python
+    name: python3
+---
+
+<!-- #region -->
+# Point Source Fluxes and Multiple Sources
+
+Once one has computed a spectral to a point source, getting the flux of that source is possible.
+In 3ML, we can obtain flux in a variety of units in a live analysis or from saved fits. There is no need to know exactly what you want to obtain at the time you do the fit.
+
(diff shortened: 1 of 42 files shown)


docs/md_docs/flux_examples.md
```diff
@@ -0,0 +1,273 @@
...
+  kernelspec:
+    display_name: Python 3
+    language: python
+    name: python3
+---
+
+<!-- #region -->
+# Point Source Fluxes and Multiple Sources
+
+Once one has computed a spectral to a point source, getting the flux of that source is possible.
```
henrikef, line 19: Nitpicky, but we can get (spatially integrated) fluxes from extended sources as well.

docs/md_docs/flux_examples.md
```diff
@@ -0,0 +1,273 @@
...
+Lets grab the result. Rememer, we can save the results to disk, so all of the following ooperations can be run at a later time without having to redo all the above steps!
+
+```python
+result = ba.results
+result.corner_plot();
+```
+
+## Computing fluxes
+
+Now we will compute fluxes. We can compute them an many different units, over any energy range also specified in any units.
```
henrikef, line 213: Suggestion: Mention explicitly that we can calculate photon fluxes (in units of photons / time/ area), energy fluxes (energy/time/area) or nuFnu fluxes ( energy^2 / time /area [????] ), and the user can pick arbitrary units for energy/area/time in each case.

henrikef (approved):
I think this is good to go, I do have some suggestions for the docs but these aren't critical.

docs/md_docs/flux_examples.md
grburgess: should we make a link to some docs on that?

henrikef: Or, just change point source to "point source and extended source" for now :D

PR #428, opened 2021-02-23
New mla

jasonfan1997:


grburgess:
This looks amazing

jasonfan1997:
Sorry I accidentally opened that merge request. The plugins for IceCube is not ready yet but is on the right track.

grburgess:
Just a note, we have switched over to a logging system. So print statements should reflect that.

We should also discuss things such as neutrino source etc.

PR #450, opened 2021-03-30
Tweak workflows

ndilalla:


.github/workflows/python-package-conda.yml
```diff
@@ -3,7 +3,7 @@ name: Python Package using Conda
 on:
   #push:
   schedule:
-    - cron: "0 20 * * *"
```
henrikef, line 6: Hm, just curious: Why did you change this one to run monthly?

henrikef (approved):

.github/workflows/python-package-conda.yml
ndilalla: Because it is failing every day and I wanted to disable it until I merge the new version I developed in a separate branch.

henrikef: Ah, makes sense.

ndilalla:
> Hm, more issues with the fermi data download. Maybe it's worth setting the fermipy tests to xfail again for now?

Yes, I think it's the right thing to do. I'll do it now.

Issue #451, opened 2021-03-31
OverflowError when positive error is inf


jasonfan1997:
**Describe the bug**
I got some customized plugin and require to run some background trials(meaning no signal). Under this case the error from the MLEResults will be inf sometime. Then when it tries to display the result, the uncertainty_formatter in io will just throw a OverflowError: cannot convert float infinity to integer 
Getting the dataframe is fine but it is the display() that tries to format the result cause this error.

**To Reproduce**
I don't have a good way to reproduce it.
But I think if you get the result table and then manually set the error(or any number to inf or NaN) in the dataframe, you will see the error when displaying it.
![Capture](https://user-images.githubusercontent.com/20786734/113085750-67fe7b00-91ae-11eb-83cb-f624759d42df.PNG)
![2](https://user-images.githubusercontent.com/20786734/113085753-67fe7b00-91ae-11eb-84b0-202fe169ce7f.PNG)

grburgess:
@jasonfan1997 this should be fixed now. If you still have issues, please reopen.

PR #458, opened 2021-04-17
fixed uncertainy formatter and added tests

grburgess:
Fixes the issues in #451

### DIFF
--- a/threeML/test/test_basic.py
+++ b/threeML/test/test_basic.py
@@ -295,20 +295,20 @@ def test_gbm_workflow():
 def test_uncertainty_formatter():
 
 
-    uncertainty_formatter(1, -1, 2)
+    assert '1.0 -2.0 +1.0' == uncertainty_formatter(1, -1, 2)
 
-    uncertainty_formatter(1e3, -1, 2)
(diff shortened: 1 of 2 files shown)


omodei (commented):
I can see that the test has passed successfully, can you also paste the output (as a comment)?

grburgess:
@omodei ah, the only difference is that now inf will be reported as inf rather than crashing the code., e.g., 
```1 -.5 + inf``` 

and nans will report zero as they always have... however, the user will be warned when either of these occur.

File: threeML/test/test_basic.py [hunk 1]
```diff
@@ -288,3 +288,27 @@ def test_gbm_workflow():
     dl_files = p.glob("*_bkg.h5")
 
     [x.unlink() for x in dl_files]
+
+
+
+
+def test_uncertainty_formatter():
```
henrikef, line 295: I guess we could explicitly test the return values here, but that might be overkill.

henrikef (approved):
Looks reasonable.

threeML/test/test_basic.py
grburgess: good point @henrikef I will add that and then merge

PR #462, opened 2021-05-14
self.restore_median_fit() added to ensure median parameters in the table describing fit 

anastasia-tsvetkova:


grburgess:
handled in #471

PR #463, opened 2021-05-14
More typing

grburgess:
Added some more typing and switching to logs


BjoernBiltzinger (approved):
Looks all good. I only fixed a few small typos.

PR #464, opened 2021-06-01
median fit restored

anastasia-tsvetkova:


anastasia-tsvetkova:
restore median fit

grburgess:
handled in #471

PR #470, opened 2021-07-08
Allow specification of parameters to be plotted in analysis_results.corner_plot

xboluna:
Allows parameters to be specified when creating the corner plot. This essentially allows creation of a custom contour plot similar to [XSPEC's plot contour](https://heasarc.gsfc.nasa.gov/xanadu/xspec/manual/XSplot.html).

e.g. in a results objects with >2 parameters, you could choose a subset of those to be plotted instead of all parameters.

### DIFF
--- a/threeML/analysis_results.py
+++ b/threeML/analysis_results.py
@@ -8,6 +8,7 @@ import math
 import os
 from builtins import map, object, range, str
 from pathlib import Path
+from typing import List, Optional, Dict
 
 import astromodels
 import astropy.units as u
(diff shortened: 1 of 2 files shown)


grburgess (approved):
Looks good to me!

grburgess:
@xboluna would you mind adding a test to make sure that this always works? 

I think you can just modify this to test the partial plotting option:

https://github.com/threeML/threeML/blob/master/threeML/test/test_analysis_results.py#L333

PR #499, opened 2021-09-08
Background deadtime

BjoernBiltzinger:
1. Fix deadtime correction for the background if the background is created from a poly fit 
2. Refactor EventList classes
3. add a np.around before type-casting the counts as np.int64, to avoid that for example 0.9999999999999999999999999 -> 0


grburgess:
I think the build is failing due to an unrelated issue. I will merge to fix this bug

PR #502, opened 2021-09-09
tiny topy in README

hamogu:
[skip CI]

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ If you find this package useful in you analysis, or the code in your own custom 
 
 
 ### Acknowledgements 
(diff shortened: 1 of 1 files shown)


grburgess:
thanks @hamogu . does that skip CI command actually work??

hamogu:
I don't know, it depends on your setup. But it *might* work in many projects (https://github.blog/changelog/2021-02-08-github-actions-skip-pull-request-and-push-workflows-with-skip-ci/), so  I thought I just try it, little harm done if it does not work.

I guess I should have written [skip ci] instead of [skip CI] thought. Not sure if it's case sensitive.

PR #506, opened 2021-09-13
Cleanup pha

grburgess:
Some cosmetics of the PHA classes to reduce technical debt


BjoernBiltzinger (approved):
Everything looks good. I really like the dataclass implementation.

PR #507, opened 2021-09-17
make warning text for infinite likelihood useful

BjoernBiltzinger:
The "Likelihood value is infinite for parameters: {trial_values}" warning in sampler_base produced not useful output because the trial_values (at least for multinest) are some ctype variables that were not printed out in a human readable way. So I changed that, to make the warning output useful.


grburgess:
@BjoernBiltzinger can you check out this test and see if merges will fix the issue?

PR #528, opened 2022-01-11
Fix initialization of 2D template function

henrikef:
This is just to keep up with the changes in astromodels.


grburgess (approved):

grburgess:
looks cool

omodei:
@henrikef can you check why all the notebooks are failing?

henrikef:
> @henrikef can you check why all the notebooks are failing?

It's the `root_numpy` issue. We could switch the jobs to run on linux, or I can try old XCode 12.4. Any preferences?

threeML/catalogs/catalog_utils.py
```diff
@@ -360,6 +358,11 @@ def _get_extended_source_from_fgl(fgl_name, catalog_entry, fix=False):
         this_shape.sigma.fix = True
         this_shape.sigma.bounds = (0, catalog_entry["model_semimajor"] / 1.36) * u.degree
 
+    elif catalog_entry["spatial_function"] == "SpatialMap":
+        
+        this_shape._load_file()
```
omodei, line 363: Isn't this already called in the constructor? I am not sure this is now needed. Please, check with DEV

henrikef: Hm, I don't think it's being called in the constructor. Maybe @grburgess can shed some light, I'm not sure I fully understand the 'defered' property: https://github.com/threeML/astromodels/blob/dev/astromodels/functions/functions_2D.py#L625

henrikef: I think you're right and this is not needed here. Might not need this pull request after all.

PR #536, opened 2022-02-11
Added the scaling factor to the background counts and errors

omodei:
Added the scaling factor to the Poisson data with Gaussian background


omodei:
@grburgess can you check our implementation (adding the scale factor to the gaussian background?) The results of the test fit slightly changed as the value of the scale factor is 0.99563578075

grburgess:
@omodei Hmm. That looks like a small numerical issue. There needs to be some logic that the value is unity UNLESS specifically specified.

PR #539, opened 2022-05-25
Switch to minuit for notebooks

henrikef:
Switch to minuit for LAT notebooks in preparation for the fermi summer school. (Fermi bottle won't have root this year). 

Tested the notebooks offline an they run fine.


henrikef:
Oops, maybe this should go into `dev`

PR #550, opened 2022-07-25
Fix response header parsing

bwgref:
Old version:
            tlmin_fchan = header["TLMIN%i" % f_chan_column_pos]

....internrept tlmin_fchan as a string, which causes a crash downstream when trying to subtract it from an int32.

Fixed by casting to an int:
            tlmin_fchan = int(header["TLMIN%i" % f_chan_column_pos])

### DIFF
--- a/threeML/io/logging.py
+++ b/threeML/io/logging.py
@@ -5,7 +5,7 @@ from contextlib import contextmanager
 from pathlib import Path
 from typing import Dict, Optional
 
-from astromodels.utils.file_utils import _get_data_file_path
+from astromodels import astromodels_config
 from astromodels.utils.logging import (LogFilter, _console_formatter,
                                        _dev_formatter, _usr_formatter,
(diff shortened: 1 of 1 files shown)


threeML/utils/OGIP/response.py
```diff
@@ -587,7 +587,7 @@ def _read_matrix(self,
         f_chan_column_pos = data.columns.names.index("F_CHAN") + 1
 
         try:
-            tlmin_fchan = header["TLMIN%i" % f_chan_column_pos]
+            tlmin_fchan = int(header["TLMIN%i" % f_chan_column_pos])
```
grburgess, line 590: Thanks for this! Would you mind using an f-string expression here? We are slowly trying to convert over to that.

bwgref:
Okay I rebased against master, but evidently introduced conflicts with logging.py. Working on it.

PR #563, opened 2022-08-19
Fix check for covariance matrix

henrikef:
The previous if statement did not catch the case when `covariance_matrix = np.ndarray(None)`


omodei (approved):
Feel free to merge when ready

PR #564, opened 2022-08-19
Backround normalization as nuisance parameters for FermipyLike

henrikef:
Added the normalization of the galactic and isotropic diffuse BG to the fermipy plugin as nuisance parameters. They are free by default.

@omodei please take a look!

One thing I don't like is how priors are handled for the bayesian analysis. For other model parameters, the user must assign a prior *before* creating the `BayesianAnalysis` object. However, the nuisance parameters are only created during that step, and we have to remember to add a prior afterwards. I wonder if we should add a default prior and/or throw a warning? Although I'm not sure how we would know inside `set_model` if we're creating a bayesian or likelihood analysis...

### DIFF
--- a/docs/md_docs/slow_execute/Fermipy_LAT.md
+++ b/docs/md_docs/slow_execute/Fermipy_LAT.md
@@ -6,7 +6,7 @@ jupyter:
       extension: .md
       format_name: markdown
       format_version: '1.3'
-      jupytext_version: 1.13.8
+      jupytext_version: 1.14.1
   kernelspec:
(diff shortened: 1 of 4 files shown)


henrikef:
NB the two failing checks are the multinest and xpec notebooks, should be unrelated.

The HAL notebook times out, I'm going to take a look but also this is not related to this pull request.

omodei (approved):

omodei:
We need to check if the doc will run with the new threeML version.

Issue #610, opened 2023-02-21
Error in following 3ML tutorial " Analyzing GRB 080916C"


poff22:
Hello, I'm following 3ML tutorial " Analyzing GRB 080916C" for my respective GRBS : GRB100826957 and GRB 130828306. I keep coming across an error on step number 8. I have attached below a screenshot of the error which happens to appear for both of the GRBs mentioned above.
Kindly assist me in how I should proceed with this.

![6](https://user-images.githubusercontent.com/113614386/220315461-8b2a1b6e-e515-4d26-a898-aac26325cdb6.png)

grburgess:
It looks like you are trying to select an interval not covered in the GBM RSP2 matrix

poff22:
Yes i understand that the interval is not being covered by the matrix but I haven't set or defined any interval for the GRB on my own. According to the code the GRB data is automatically donwloaded from GBM burst catalog and so the intervals are predefined.
So i dont understand why this error appears .Do you have any suggestions as to what could be causing it and how I should proceed?

grburgess:
The interval must have been defined. If you can provide a formatted code example that would be more useful.

poff22:
[GRB_100826957_GBM_analysis.txt](https://github.com/threeML/threeML/files/10811368/GRB_100826957_GBM_analysis.txt)

poff22:
I have shared the code I am following above. Looking forward to your response.

grburgess:
Please include the code with formatting within the issue.

poff22:
[terminal.txt](https://github.com/threeML/threeML/files/10836058/terminal.txt)
[response.txt](https://github.com/threeML/threeML/files/10836061/response.txt)


I have attached above two files. FIle that goes by the name of terminal consists of the output we get when we run the code on the temrinal.
respont file is the code found in threeML packages - the error appears to be found in line 1115. 

I hope these are the files you are referring to - if not then please guide me as to which file i sould share.

grburgess:
Please follow the issue templates. Code should be included as formatted markdown. Not as "txt" files

henrikef:
@poff22 

It is much safer and easier to read if you include your code and log file in the issue text in between three ` signs. It will show up like so, properly formatted:

```python
#python code goes here
import threeML
```

You do not need to upload threeML code here, only code or scripts you yourself wrote and the output you get, so that we can duplicate the issue.

Now in your case, I saw that there's a bug in printing the error message. Fixing that, I get the following error:

```
[...]

PR #611, opened 2023-02-27
Fixing error message in OGIP/response.py

henrikef:
See disussion in #610


henrikef:
I think some of these are just failing because the astromodels on conda is behind `dev`. 

``` 
TypeError: get_point_source_fluxes() got an unexpected keyword argument 'stokes'
``` 

should be fixed on astromodels dev.

grburgess:
Looks fine to me! Thanks for grabbing that

PR #612, opened 2023-03-09
Fermipylike small updates

henrikef:
Some small changes and fixes.

* Sanitize `4FGL...`  source names to `x4FGL...` instead of `_4FGL`. Matplotlib does not like to display labels starting with an underscore.
* Fix pivot energy for logP spectra read in from a fermipy saved ROI
* Fix extended source read in from fermipy saved ROI

To do: Check docs


grburgess:
@omodei does anyone know if this I ready to go now?

PR #623, opened 2023-07-21
Nuatilus

grburgess:
add the nautilus sampler.


johannesulf:
This implementation looks great! Please don't hesitate to reach out to me if you have any general questions about nautilus (https://github.com/johannesulf/nautilus).

grburgess:
Thanks! Yes it works very nicely actually. Unfortunately, the way our likelihoods are implemented prevents it from running with multiprocessing, but maybe we can fix that at some point

grburgess:
@johannesulf it is just an issue with being able to pickle the analysis. It works in some cases, but not all.

PR #628, opened 2023-12-28
Fixing issue with mutable default value in several class defintions.

torresramiro350:
When trying to install `threeML`  with python 3.11, I got several of these errors when launching `threeML`  where a mutable default is set several class definitions.

 Here's one of the several that I got when I tried `import threeML`
```
` ValueError: mutable default <class 'threeML.config.catalog_structure.InstrumentCatalog'> for field Fermi is not allowed: use default_factory
```
The fix is to use `field(default_factory=InstrumentCatalog)` when setting the default value.


ndilalla:
Thanks Ramiro! Pull request manually merged in commit [0dffe92](https://github.com/threeML/threeML/commit/0dffe924d2249234063051268659a0705e13ef75)

PR #646, opened 2025-05-21
Send logs to stderr by default

israelmcmc:
Python's builtin logging library prints by default to stdout, not stderr. This is the convention since it allows you to easily separate the applications output from the logs. This is, for example, useful when working on larger pipelines, where the logs and output are usually redirected to different places.

This is related to this other issue in Rich, which threeML uses:
https://github.com/Textualize/rich/issues/2022

While the maintainer hasn't changed the default to stderr, the thread has multiple arguments from various developers of why this should be the default. ThreeML uses a custom `Console` when calling `RichHandler`, so the Rich default doesn't matter, but I thought it was still related.

In principle, just adding the flag `stderr=True` during the `Console` instantiation should had been enough, but the log were still being outputted to stdout.

The problem was that the main `__init__` is importing everything from `astromodels`, including the astromodel's `setup_logger`, which was shadowing theeML's `setup_logger`.

Fixing the order of the imports revealed a circular import, which was preventing importing threeML at all. The `io/logging` moduled needed the `config` module to configure the logger, and the `config` module needed `io/logging` to create its own logs. I patched this by moving the `setup_logger` import in `config` to local import within the functions where it was needed.

While this patch worked for now to achieve the initial goal of using logging to stderr by default, a circular import is usually a symptom telling you that there's something in the code design that needs to changed. In this case, I think the way to go to solve this circular import is to not configure the logging within 3ML at all. This is also the recommended way to use the logging library. I opened the issue #647 about this, which I can address in a separate PR. 

The equivalent PR for astromodels is: https://github.com/threeML/astromodels/pull/217

### DIFF
--- a/threeML/__init__.py
+++ b/threeML/__init__.py
@@ -21,10 +21,10 @@ except ImportError:
 
 from pathlib import Path
 
-from threeML.io.logging import setup_logger
-
 # Import everything from astromodels
 from astromodels import *
+
+from .io.logging import setup_logger
 
 from .config import (
     threeML_config,
--- a/threeML/config/config_utils.py
+++ b/threeML/config/config_utils.py
@@ -5,13 +5,15 @@ from typing import Any, Dict, Optional
 from omegaconf import OmegaConf
 from omegaconf.dictconfig import DictConfig
 from rich.tree import Tree
-from threeML.io.logging import setup_logger
 from threeML.io.package_data import get_path_of_user_config
 
 from .config import threeML_config
 
-log = setup_logger(__name__)
-
+# FIXME: These lines were moved to local imports withing functions since
+#  they were causing a circular import. The config module needs the logging setup,
+#  and the logging setup needs the config module.
+#from threeML.io.logging import setup_logger
+#log = setup_logger(__name__)
 
 def recurse_dict(d, tree):
 
@@ -64,6 +66,8 @@ def show_configuration(sub_menu: Optional[str] = None):
 
             msg = f"{sub_menu} is not in the threeml configuration"
 
+            from threeML.io.logging import setup_logger
+            log = setup_logger(__name__)
             log.error(msg)
 
             raise AssertionError(msg)
@@ -111,6 +115,10 @@ def get_value(name, user_value, par_type, config_value):
     :param config_value: value in config
     :returns: parameter value
     """
+
+    from threeML.io.logging import setup_logger
+    log = setup_logger(__name__)
+
     if user_value is not None:
         value = user_value
     else:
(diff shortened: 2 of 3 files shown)


israelmcmc:
🤔 I think the failing tests are not related to my changes, are they? The `dev` branch shows all tests passed, but the job that is failing here was skipped during the `dev` tests: https://github.com/threeML/threeML/actions/runs/15124602790 . "Build and verify with conde" was skipped in favor of "Publish with conda". I'm not sure what's exactly the difference.

Or at least that's what it seems to me, I'm not very familiar with the tests suite. Any ideas? I at least verified that all tests passed in my machine.

ndilalla:
> 🤔 I think the failing tests are not related to my changes, are they? The `dev` branch shows all tests passed, but the job that is failing here was skipped during the `dev` tests: https://github.com/threeML/threeML/actions/runs/15124602790 . "Build and verify with conde" was skipped in favor of "Publish with conda". I'm not sure what's exactly the difference.
> 
> Or at least that's what it seems to me, I'm not very familiar with the tests suite. Any ideas? I at least verified that all tests passed in my machine.

This looks more like a temporary glitch of the internet connection as it fails to download the catalog files. I roll them back to see if this solves the issue. They don't look related to this PR.

israelmcmc:
Thanks @ndilalla! Most are passing now, except for "test_dynesty_nested" on (macos13, python 3.9). It's passing in both (macos13, python 3.11) and (macos-latest, python 3.9). It also doesn't seem related to the changes here. 

Is is possible that the results of that test are random and that the test values tolerance is too small to account for the spread?

ndilalla:
> Thanks @ndilalla! Most are passing now, except for "test_dynesty_nested" on (macos13, python 3.9). It's passing in both (macos13, python 3.11) and (macos-latest, python 3.9). It also doesn't seem related to the changes here.
> 
> Is is possible that the results of that test are random and that the test values tolerance is too small to account for the spread?

It shouldn't be random, but you never know! By looking at the fit results in the log, it looks like the fit didn't converge properly:
`bn090217206.spectrum.main.Powerlaw.K        2.9392650642395557 -0.8 +0`
I restarted again this job, I am positive this will solve this issue.

israelmcmc:
Thanks @ndilalla, that fixed it indeed.

israelmcmc:
Thanks, @ndilalla. Yeah, I think so. At least I'm not planning to add anything else to this PR. I can work on #647 on a separate PR, if that OK with you and @omodei.

I'd recommend merging this one together with https://github.com/threeML/astromodels/pull/217

PR #654, opened 2025-10-14
Support for Dynesty v3.0.0

PreisTo:
Change the `NestedSampler` and `DynamicNestedSampler` `setup()` to be compatible with the latest major `dynesty` release which involved major refactoring.

Default values are the same as used by `dynesty`.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -144,6 +144,9 @@ jobs:
                   - environment: ci/environment_fermi.yml
                   - environment: ci/environment_macos-latest.yml
                     os: macos-latest
+                  - environment: ci/environment_macos-latest.yml
+                    os: macos-15-intel
+                    python-version: 3.9
         runs-on: ${{ matrix.os }}
         steps:
             - name: Checkout
--- a/.github/workflows/conda_build.yml
+++ b/.github/workflows/conda_build.yml
@@ -120,6 +120,9 @@ jobs:
                   - environment: ci/environment_fermi.yml
                   - environment: ci/environment_macos-latest.yml
                     os: macos-latest
+                  - environment: ci/environment_macos-latest.yml
+                    os: macos-15-intel
+                    python-version: 3.9
         steps:
             - name: Checkout
               uses: actions/checkout@v4


ndilalla:
Thanks @PreisTo! I see that unfortunately `dynesty 3.0.0` was released on [conda](https://anaconda.org/conda-forge/dynesty/files) with support for `python>=3.10`, even though on `pypi` and its docs it says `python>=3.8`. Since the update looks not backward compatible, I guess we have two possible ways to solve the issues on `python 3.9`:
- Not installing and testing `dynesty` on `python 3.9`
- Forcing the installation of `dynesty 3.0.0` on the conda environment with `pip` and then testing it as usual
- Checking runtime the version of `dynesty` installed and change our code to support older versions

The first two are definitely easier to implement, and I would probably go for the second one. I don't know how feasible is to keep supporting older versions of `dynesty`, as changes look pretty major. Any thoughts?

ndilalla:
Also, I see that the latest version of `h5py` (3.15, available so far on pypi only) fails the installation on `macos-13`. I wonder what the issue is...

PreisTo:
> Thanks @PreisTo! I see that unfortunately `dynesty 3.0.0` was released on [conda](https://anaconda.org/conda-forge/dynesty/files) with support for `python>=3.10`, even though on `pypi` and its docs it says `python>=3.8`. Since the update looks not backward compatible, I guess we have two possible ways to solve the issues on `python 3.9`:
> 
>     * Not installing and testing `dynesty` on `python 3.9`
> 
>     * Forcing the installation of `dynesty 3.0.0` on the conda environment with `pip` and then testing it as usual
> 
>     * Checking runtime the version of `dynesty` installed and change our code to support older versions
> 
> 
> The first two are definitely easier to implement, and I would probably go for the second one. I don't know how feasible is to keep supporting older versions of `dynesty`, as changes look pretty major. Any thoughts?

I actually think option 3 is the easiest to implement (already did and works fine for the `dynesty` versions I've tested using a conda environment + conda & pip install), by simply passing all the keyword arguments as `**kwargs` and simply pass them along to the `dynesty` objects. This would come at the cost of the user having to check the docs of their current `dynesty` implementation and what keywords are available for the wanted sampler.

ndilalla:
> @ndilalla for better compatibility and "future-proofness" I think that 3ML should not interfere too much with the arguments of the samplers.
> 
> My suggestion would be to allow to pass `**kwargs` to the `BayesianAnalysis.sample()` function which then get passed further (see the changes in that file). Currently this does not mess with any other sampler, however to be consistent in how to setup and sample one could argue to modify the other ones in the same way. What are your thoughts on this one?

I agree 100%. Ideally I would like to leave a **very** minimal set of frequently used arguments explicit, like `n_live_points`  and a few others. I think `dynesty` is the only sampler which has tons of exposed keywords, while the others have just a few, plus the final `**kwargs`.  If you agree, I would change `dynesty` to be similar to the other samplers. We need then to make sure to include the links to the sampler documentation, so that it's easy to check which other keywords are accepted by the `setup()` method.

PreisTo:
> > @ndilalla for better compatibility and "future-proofness" I think that 3ML should not interfere too much with the arguments of the samplers.
> > My suggestion would be to allow to pass `**kwargs` to the `BayesianAnalysis.sample()` function which then get passed further (see the changes in that file). Currently this does not mess with any other sampler, however to be consistent in how to setup and sample one could argue to modify the other ones in the same way. What are your thoughts on this one?
> 
> I agree 100%. Ideally I would like to leave a **very** minimal set of frequently used arguments explicit, like `n_live_points` and a few others. I think `dynesty` is the only sampler which has tons of exposed keywords, while the others have just a few, plus the final `**kwargs`. If you agree, I would change `dynesty` to be similar to the other samplers. We need then to make sure to include the links to the sampler documentation, so that it's easy to check which other keywords are accepted by the `setup()` method.

Just to make sure I do not misunderstand you: I'll modify this PR such that there are, similar to e.g. MultiNest, just some keywords like `n_live_points` and other frequently used options, set them to the `dynesty` default values and pass the rest as part of the `**kwargs`

ndilalla:
> > > @ndilalla for better compatibility and "future-proofness" I think that 3ML should not interfere too much with the arguments of the samplers.
> > > My suggestion would be to allow to pass `**kwargs` to the `BayesianAnalysis.sample()` function which then get passed further (see the changes in that file). Currently this does not mess with any other sampler, however to be consistent in how to setup and sample one could argue to modify the other ones in the same way. What are your thoughts on this one?
> > 
> > 
> > I agree 100%. Ideally I would like to leave a **very** minimal set of frequently used arguments explicit, like `n_live_points` and a few others. I think `dynesty` is the only sampler which has tons of exposed keywords, while the others have just a few, plus the final `**kwargs`. If you agree, I would change `dynesty` to be similar to the other samplers. We need then to make sure to include the links to the sampler documentation, so that it's easy to check which other keywords are accepted by the `setup()` method.
> 
> Just to make sure I do not misunderstand you: I'll modify this PR such that there are, similar to e.g. MultiNest, just some keywords like `n_live_points` and other frequently used options, set them to the `dynesty` default values and pass the rest as part of the `**kwargs`

Exactly!

ci/environment_fermi.yml
```diff
@@ -52,3 +53,4 @@ dependencies:
   - rich
   - joblib
   - fermipy
+  - llvmlite
```
ndilalla: `llvmlite` is available on `conda-forge` too, no need to add `numba` channel. Anyway I am not sure we should add this, it's not our direct dependency, rather something required by `numba`. Am I wrong?

ci/environment_macos-latest.yml
```diff
@@ -47,4 +48,5 @@ dependencies:
   - h5py
   - rich
   - joblib
-  - fermipy
\ No newline at end of file
+  - fermipy
+  - llvmlite
```
ndilalla: Same here as above

PreisTo:
@ndilalla I've removed the `n_effective_init` statment (just included it bcs it was explicitly used in a previous test) so that solved that issue.

However I the `macos-15-intel` issue remains. I did some minor digging into it and noticed that the failing versions on `macos-15-intel` use the beta release of `numba==0.63.0b1`. where as the `pip` and `conda` versions not failing on the os use an older release. So my guess is that this might be the causing issue, however I could not find any further info on why its failing.

pip-python 3.11: numba-0.63.0b1-cp311-cp311-macosx_10_9_universal2.whl.metadata
pip-python 3.10: numba-0.63.0b1-cp310-cp310-macosx_10_9_universal2.whl.metadata

in comparison
pip-python 3.9: numba-0.60.0-cp39-cp39-macosx_10_9_x86_64.whl.metadata (non failing)

or
conda-python 3.11: numba                     0.62.1          py311hf901296_0    conda-forge


I do not know about an option to enforce `pip` to use only stable releases. What do you think is the best way to go?

PR #662, opened 2026-01-26
Fix nautilus sampler

PreisTo:
Adds test for `nautilus` + minimize exposure of setup keywords, similar to #654


threeML/bayesian/nautilus_sampler.py
```diff
@@ -8,24 +7,10 @@
 from threeML.config.config import threeML_config
 from threeML.io.logging import setup_logger
 
-
-def capture_arguments(func, *args, **kwargs):
```
ndilalla: @PreisTo Is this function no longer needed?

PreisTo: This seems to have never been used - there is no reference of this function neither in 3ml nor in nautilus and nothing mentioned in the docs

Issue #663, opened 2026-02-11
PluginPrototype not being garbage-collected when used with JointLikelihood fitting


jdbuhler:
**Describe the bug**
We (cosipy) have created a PluginPrototype derivative, COSILike, that allocates a large array internally at __init__ time.  When a  COSILike object is used as part of a DataList that is then used to initialize a JointLikelihood model that is used for fitting, the object is not garbage-collected even after it, the DataList, and the JointLikelihood all go out of scope.  Consequently, if we wish to perform fitting multiple times in a loop with a slightly modified COSILike each time, we rapidly run out of memory.

A minimal script to reproduce the problem using only threeML, without any cosipy dependencies, is shown below.  Here's what I get (cut off after a few iterations).  The 1 GB allocated in the PluginPrototype derivative in each iteration is never freed.

```python
13:24:32 WARNING   The naima package is not available. Models    functions.py:43
                  that depend on it will not be available
         WARNING   The GSL library or the pygsl wrapper cannot   functions.py:65
                  be loaded. Models that depend on it will not
                  be available.
         WARNING   The ebltable package is not available.       absorption.py:33
                  Models that depend on it will not be
                  available
13:24:33 INFO      Starting 3ML!                                  __init__.py:44
         WARNING   WARNINGs here are NOT errors                   __init__.py:45
         WARNING   but are inform you about optional packages     __init__.py:46
                  that can be installed
         WARNING    to disable these messages, turn off           __init__.py:47
                  start_warning in your config file
         WARNING   no display variable set. using backend for     __init__.py:53
                  graphics without display (agg)
         WARNING   ROOT minimizer not available             minimization.py:1208
         WARNING   Multinest minimizer not available        minimization.py:1218
         WARNING   PyGMO is not available                   minimization.py:1228
13:24:34 WARNING   The cthreeML package is not installed. You     __init__.py:95
                  will not be able to use plugins which require
                  the C/C++ interface (currently HAWC)
         WARNING   Could not import plugin FermiLATLike.py. Do   __init__.py:136
                  you have the relative instrument software
                  installed and configured?
         WARNING   Could not import plugin HAWCLike.py. Do you   __init__.py:136
                  have the relative instrument software
                  installed and configured?
         WARNING   No fermitools installed           lat_transient_builder.py:44
ITERATION 0
RAM usage: 364.51 MB
         INFO      set the minimizer to minuit           joint_likelihood.py:994
[...]

PR #664, opened 2026-02-17
Using weakref instead of strong callback links.

ndilalla:
Implemented a reference-retention fix for issue #663 by removing strong callback links from optimization/sampling backends to analysis objects.

### Root cause
* The objective/posterior callbacks were passed as bound methods (self.method).
* Backends (notably iminuit, and similarly MCMC samplers) keep those callables, which can retain the full analysis object graph (JointLikelihood / SamplerBase -> DataList -> plugins), preventing timely garbage collection.

### Fix and changes
* Added a weakref-backed callable proxy in ``threeML/classicMLE/joint_likelihood.py``: ``_WeakMinusLogLikeProxy``
* Replaced all minimizer callback wiring to use that proxy instead of bound methods.
* Added weakref-backed posterior callable in ``threeML/bayesian/sampler_base.py``: ``get_posterior_proxy()``
* Updated unit-cube posterior builder (``_construct_unitcube_posterior``) so loglike and prior closures use weakrefs instead of strongly capturing self.
Updated samplers that explicitly used bound methods: ``emcee`` and ``zeus``.
* Added a test ``threeML/test/test_gc_issue.py``

----
📚 Documentation preview 📚: https://threeml--664.org.readthedocs.build/en/664/

### DIFF
--- a/threeML/bayesian/emcee_sampler.py
+++ b/threeML/bayesian/emcee_sampler.py
@@ -98,12 +98,12 @@ class EmceeSampler(MCMCSampler):
                 view = c[:]
 
                 sampler = emcee.EnsembleSampler(
-                    self._n_walkers, n_dim, self.get_posterior, pool=view
+                    self._n_walkers, n_dim, self.get_posterior_proxy(), pool=view
                 )
 
             else:
                 sampler = emcee.EnsembleSampler(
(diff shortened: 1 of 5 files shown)


PreisTo (commented):
@ndilalla looks really good to me :)
Should we add some docs explaining how and when to use the method used in the tests?

Issue #647, opened 2025-05-21
Change how threeML configures logging (or rather, do not configure logging)


israelmcmc:
ThreeML and astromodels currently configure the loggers through the setup_logger functions:
https://github.com/threeML/threeML/blob/7a16580d9d5ed57166e3b1eec3d4fccd3eeef1eb/threeML/io/logging.py#L392
and 
https://github.com/threeML/astromodels/blob/df43e68c147c36fd760a589d6223ddc35fb27cbe/astromodels/utils/logging.py#L196

ThreeML and astromodels are general purpose libraries that are imported within larger applications. The Python documentation recommends that libraries should not configure the logging format, and explicitely to not add any handler: 

> Note: It is strongly advised that you do not add any handlers other than [NullHandler](https://docs.python.org/3/library/logging.handlers.html#logging.NullHandler) to your library’s loggers. This is because the configuration of handlers is the prerogative of the application developer who uses your library. The application developer knows their target audience and what handlers are most appropriate for their application: if you add handlers ‘under the hood’, you might well interfere with their ability to carry out unit tests and deliver logs which suit their requirements.

I agree with this philosopy in general, but it became more concrete while working on PR #646 as part of the work to use threeML in the COSI pipeline.

I can work on this unless there's an objection. I would still leave an option to use the current configuration in case an application or user wants that formatting and handlers, but all `setup_logger(__name__)` within the library would be changed to `logging.getLogger(__name__)`.

PreisTo:
Closing this as solved with merged #672 for threeML v2.6.1

PR #672, opened 2026-03-25
Use logging.getLogger(__name__) instead of setup_logger.

israelmcmc:
This PR addresses #647

The philosophy is that 3ML is a library, which can be imported by larger libraries or by other application. As such, it should not handle it's own logs, only produce them and delegate the logging parent logging.

The main change is to follow best practices and have a `logging.getLogger(__name__)` at the beginning of each file instead of 3Ml customs `setup_logger`. 

The customization of the logger should only be done `__main__` so it can propagate to the loggers of all libraries. If the user want to recover the previous logging behavior (e.g. formatting, time rotatin file handlers) they can achieve this by explicitly calling 3ML's `setup_logger` in their `__main__`. Example: https://gist.github.com/israelmcmc/6bebe314b9d7bddcf6fecc0240926f96

In order to allow the user to decide how the startup_warning are formatting and where they go, there's now a function called `log_threeml_startup_warnings` (and `log_threeml_startup_warnings` from astromodels), that accepts the logger that will handle these logs. This was necessary because otherwise they would be printed during the initial import, before we could even call `setup_logger`. However, this also removed the need of a config file to prevent these warning, which is a plus imho. 

Lastly, I also moved `get_path_of_user_config` to fix a previously hidden circular import (the config needed the logging to log, and the logging module needed the config module to know what to do log!).

Astromodels companion: https://github.com/threeML/astromodels/pull/260

### DIFF
--- a/docs/md_docs/fast_execute/configuration.md
+++ b/docs/md_docs/fast_execute/configuration.md
@@ -66,7 +66,6 @@ plugins:
 ```yaml
 logging:
   developer: on
-  startup_warnings: off
 ```
 
 ```python
@@ -84,7 +83,6 @@ logging:
   path: ${env:USER_LOG_DIR}
   developer: off
   console: off
-  startup_warnings: off
 ```
 
 where ```USER_LOG_DIR``` could be a directory where you are allowed to write to disk. At run time, we convert this variable to a path. However, any value can be set. Thus, you can make an adaptable configuration that mutates based on your local workstation. 
--- a/docs/md_docs/fast_execute/logging.md
+++ b/docs/md_docs/fast_execute/logging.md
@@ -16,7 +16,40 @@ jupyter:
 <!-- #region -->
 # Logging and Verbosity
 
-We have recently switched to reporting information to the user via logging. Logging in 3ML occurs in three different ways:
+We use the [`logging`](https://docs.python.org/3/library/logging.html) library to manage all logs. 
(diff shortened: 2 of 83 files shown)


israelmcmc:
There's an unit test failing about a missing `_ra`. I tracked it down to his query not working:
https://heasarc.gsfc.nasa.gov/xamin/vo/cone?showoffsets&table=fermille&DEC=0&RA=0&SR=300&VERB=3

Return 
```
<INFO name="Error" value="Query Error: Unable to execute query: select __row, trigger_name, name, ra, dec, lii, bii, time, end_time, trigger_time, trigger_type, version, degrees(2*asin(sqrt(power(sin((radians( dec ) - 0.0)/2),2) +cos(radians( dec ))*cos(0.0)*power(sin((radians( ra ) - 0.0)/2),2))))*60 as "_offset" from fermille where ( ( ((1 = 1)) ) ) limit 10001"/>
```

Which causes a `None` [here](https://github.com/threeML/threeML/blob/749e78c21f694a0c41fbd78c1b53a076bcd894bf/threeML/catalogs/VirtualObservatoryCatalog.py#L105) and prevent `_ra` from being initialized below. 

Any ideas @ndilalla @omodei? It guess it's unrelated to this PR, just an issue with the HEASARC server.

israelmcmc:
I'm rerunning all jobs. Someone at HEASARC told me that the issue with the cone search should be resolved now.

israelmcmc:
Thanks @ndilalla!

PR #684, opened 2026-07-22
Gammapy plugin

PreisTo:
## Summary

- added `GammapyLike` to the supported  plugins for `display_spectrum_model_counts`
- fixed the core functionality of `corner_plot_cc` for `chainconsumer>=1`
- fix failing `plot_spectra` if `use_components` and `total` is supplied
- some minor docs stuff

----
📚 Documentation preview 📚: https://threeml--684.org.readthedocs.build/en/684/

### DIFF
--- a/ci/environment.yml
+++ b/ci/environment.yml
@@ -48,3 +48,4 @@ dependencies:
   - joblib
   - pip:
     - autoemcee
+    - chainconsumer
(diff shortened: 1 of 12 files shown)


PreisTo:
`FittedPointSourceSpectralHandler` fails in in `_build_propagated_function`  in case there is a composite function with one of the components having 0 free parameters 

Need a way to combine it to another part of the composite and call them simultaneously so that there is at least one free parameter

ndilalla (approved):

PR #686, opened 2026-08-13
[FEAT]: More logging info discarded samples

PreisTo:
We currently do not provide a lot information when initializing a MLEResult and a lot of "illegal" samples are created in the process, as usually the bounds are to tight.

This PR adds some more information on the discarded samples so the user can better debug the model - this is from the added test (both parameters converged against the lower bound):
```bash
  WARNING UserWarning: 73.6% of samples have been thrown away because they failed the constraints on the parameters. This result might not be suitable for error propagation. Enlarge the boundaries until you lose less than 1% of the samples.
  Discarded samples in excluded parameter space (3679 samples):
    K: mean=105, std=3.819e-08, bounds=[105, 115]
    index: mean=-2.8, std=1.2e-09, bounds=[-2.8, -2.6]
```

----
📚 Documentation preview 📚: https://threeml--686.org.readthedocs.build/en/686/

### DIFF
--- a/threeML/analysis_results.py
+++ b/threeML/analysis_results.py
@@ -5,11 +5,11 @@ import inspect
 import logging
 import math
 import os
-from builtins import map, object, range, str
+import warnings
+from builtins import map, range, str
 from importlib.util import find_spec
 from pathlib import Path
 from typing import Dict, List, Optional
-import warnings
 
 import astromodels
 import astropy.units as u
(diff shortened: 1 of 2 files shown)


ndilalla:
Absolutely love this! Thanks @PreisTo :100:

ndilalla (approved):

PreisTo:
@ndilalla that was Seonghyeon Yu's idea :)

PR #689, opened 2026-08-31
Fix Docs CI 

PreisTo:
----
📚 Documentation preview 📚: https://threeml--689.org.readthedocs.build/en/689/

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -142,6 +142,11 @@ jobs:
           pip install -e .
         env:
           ISDEV: ${{ github.ref == 'refs/heads/dev' || github.base_ref == 'dev' || github.head_ref == 'dev' }}
+
+      - name: Conda list
+        shell: bash -l {0}
+        run: |
+            mamba list
       - name: Execute the notebooks
         shell: bash -l {0}
         run: |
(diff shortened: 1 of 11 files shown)


PreisTo:
@ndilalla I fixed the channel conflict between the setup and the env files + cleaned up a bit the setup procedure + added a "list" step + enforced the threeml channel in the env files for astromodels

Now all docs are actually using the threeml astromodels version :)

.github/workflows/docs.yml
```diff
@@ -26,17 +26,14 @@ jobs:
...
-             use-only-tar-bz2: false
-             miniforge-version: latest
+          auto-update-conda: true
+          auto-activate: false
+          miniforge-version: latest
+          mamba-version: "*"
+          activate-environment: test_env
+          python-version: 3.12
+          environment-file: ci/environment.yml 
+          use-only-tar-bz2: true
```
ndilalla: What's the reason we are setting ` use-only-tar-bz2` to true (here and after)? If I understand correctly, this way we can't install packages with `.conda` format. I am surprised it even works.

ndilalla:
Thanks @PreisTo ! I left a comment above. Also, I see that `test_against_xspec.yml` and `build_and_test.yml` also have some workflows using `mamba-version: "2.5"`. Should we change those as well?

.github/workflows/docs.yml
PreisTo: That was solely due to setup-miniconda stating
> For caching to work properly, you will need to set the use-only-tar-bz2 option to true.

But honestly I don't think we need it with `mamba`

ndilalla (approved):

